PR #13, opened 2021-07-09
Add cli

hrshdhgd:
- JSONLoader
- RDFLoader
- YAMLLoader
- JSONDumper
- RDFDumper
- YAMLDumper

### DIFF
--- a/.github/workflows/pr-test.yaml
+++ b/.github/workflows/pr-test.yaml
@@ -10,7 +10,7 @@ jobs:
 
     runs-on: ubuntu-latest
     strategy:
-      python-version: [3.7.1, 3.8, 3.9]
+    - python-version: [3.7.1, 3.8, 3.9]
 
     steps:
     - uses: actions/checkout@v2
--- a/.github/workflows/pr-test.yaml
+++ b/.github/workflows/pr-test.yaml
@@ -10,7 +10,7 @@ jobs:
 
     runs-on: ubuntu-latest
     strategy:
-    - python-version: [3.7.1, 3.8, 3.9]
+      python-version: [3.7.1, 3.8, 3.9]
 
     steps:
     - uses: actions/checkout@v2
--- a/.github/workflows/pr-test.yaml
+++ b/.github/workflows/pr-test.yaml
@@ -9,8 +9,9 @@ jobs:
   build-pipenv:
 
     runs-on: ubuntu-latest
-    strategy:
-      python-version: [3.7.1, 3.8, 3.9]
+    strategy:      
+      matrix:
+        python-version: [3.7.1, 3.8, 3.9]
 
     steps:
     - uses: actions/checkout@v2
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,6 +1,50 @@
 CHANGES
 =======
 
+v1.0.10
+-------
+
+* Fix an omission on inline generation
+
+v1.0.9
+------
+
+* Sync generated files
+* Add MissingRequiredField to YAMLRoot
+* Phase 1 in adding MissingRequiredField intercept
+* Update main.yaml
+
+v1.0.8
+------
+
+* Fix import error
+* Add sequential dependencies to build process
+* Update Pipfile
+
+v1.0.7
+------
+
+* Automatically generated requirements.txt and requirements-dev.txt
+* Fix for issue #6
+* Automatically generated requirements.txt and requirements-dev.txt
+* Fix issue #3
+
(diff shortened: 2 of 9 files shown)


ChangeLog
```diff
@@ -1,19 +1,63 @@
 CHANGES
 =======
 
+v1.0.10
```
cmungall, line 4: curious why this is coming in in your PR @hrshdhgd ?

linkml_runtime/dumpers/rdf_dumper.py
```diff
@@ -81,7 +81,77 @@ def dumps(self, element: YAMLRoot, contexts: CONTEXTS_PARAM_TYPE = None, fmt: Op
...
         """
         return self.as_rdf_graph(element, contexts).serialize(format=fmt).decode()
+
+@click.group()
+def cli():
+    pass
+
+@cli.command('as_rdf_graph')
+@click.option('-e', '--element', help='LinkML object to be emitted.')
+@click.option('-c', '--contexts', help='JSON-LD context(s)')
```
cmungall, line 95: having `(s)` indicates this can be multivalued but this seems a single valued option?

ChangeLog
hrshdhgd: I noticed that too! Whenever I do `python setup.py install` when first setting up my virtual environment, this one and the 'AUTHORS' file get generated.

linkml_runtime/loaders/json_loader.py
```diff
@@ -25,3 +25,28 @@ def loader(data: Union[str, dict], _: FileInfo) -> Optional[Dict]:
...
+    """Loads data from source \n
+    Args: \n
+        input (str): Input data to load \n
+        target (str): Target Class \n
+        base_dir (str): Base directory \n
+        metadata (str): Metadata \n
+    Returns: \n
+        YAMLRoot
+    """
+    JSONLoader.load(source=input, target_class=target, base_dir=base_dir, metadata=metadata)
```
cmungall, line 49: have you tried this?

The signature of load is

```python
   def load(self, source: Union[str, dict, TextIO], target_class: Type[YAMLRoot], *, base_dir: Optional[str] = None,
 ```

but your target_class is a string - does it get auto-converted?

File: linkml_runtime/dumpers/rdf_dumper.py (hunk 2 above)
Comment (hrshdhgd) [reply linkml_runtime/dumpers/rdf_dumper.py:95]: I missed it, sorry about that. Does `@click.argument('contexts', required=True')` make sense?

File: linkml_runtime/loaders/json_loader.py (hunk 3 above)
Comment (hrshdhgd) [reply linkml_runtime/loaders/json_loader.py:49]: No it does not .Fixed it.

File: ChangeLog (hunk 1 above)
Comment (hsolbrig) [reply ChangeLog:4]: A product of the PBR loader -- pretty useful

File: .github/workflows/pr-test.yaml [hunk 4]
```diff
@@ -9,8 +9,9 @@ jobs:
   build-pipenv:
 
     runs-on: ubuntu-latest
-    strategy:
-      python-version: [3.7.1, 3.8, 3.9]
+    strategy:      
+      matrix:
+        python-version: [3.7.1, 3.8, 3.9]
```
cmungall, line 14: wondering why this is in this PR - seems like a separate concern?

cmungall (changes requested):
Apologies, I gave you bad guidance when I specced this out

in fact it is not useful to run "load" in isolation from the command line, as the end result of load is an in-memory python object

And I don't think "dump" works, as the input for dump is also an in-memory object

In fact I think we want something more like sssom convert here. Make sense?

sorry for the misidirection

.github/workflows/pr-test.yaml
hrshdhgd: This was throwing an error while running tests. So this seemed to be a quick fix. not sure if it is supposed to be this way.

hrshdhgd:
> Apologies, I gave you bad guidance when I specced this out
> 
> in fact it is not useful to run "load" in isolation from the command line, as the end result of load is an in-memory python object
> 
> And I don't think "dump" works, as the input for dump is also an in-memory object
> 
> In fact I think we want something more like sssom convert here. Make sense?
> 
> sorry for the misidirection

So if I understand this correctly, you want the CLI to load and dump in a function? e.g. `json_to_rdf` or `rdf_to_yaml` etc. ?

cmungall:
closed in favor of different approach: #26.

PR #23, opened 2021-07-29
replace hyphen with underscore when underscoring names

cmungall:


gaurav:
Note that this only replaces whitespace, `,` and `-` with underscores -- maybe we should modify this regular expression so that it replaces all non-word characters with underscores (i.e. `\W` from https://docs.python.org/3/library/re.html)?

PR #26, opened 2021-08-09
added dataconvert CLI tool

cmungall:
This adds a command line tool `dataconvert` (open to other names...) which takes as input instance data and uses linkml-runtime to (1) instantiate in-memory objects (2) re-serialize


cmungall:
the main linkml repo is a better place for this

PR #35, opened 2021-09-08
add a convenience method for retrieving model elements by the prefix …

sierra-moxon:
…of the supplied curie see biolink-model-toolkit issue [#61](https://github.com/biolink/biolink-model-toolkit/issues/61)


cmungall (commented):
hmm, the change in signature from all_X to all_Xs will break some downstream code, the notebooks (which we should run tests over but we don't)

Would it be possible to make the get_element_by_prefix change it's own PR, and we can discuss changes to the signature separately?

maybe we could retain the old signatures as a deprecated "redirect"..? either way, a separate PR

sierra-moxon:
ah, good point.  yep, will extract just the method addition (now PR #36) and do the refactoring in another PR with deprecation methods.

PR #36, opened 2021-09-09
function to return schema elements by id_prefixes 

sierra-moxon:
- remove extra refactoring from PR #35

### DIFF
--- a/linkml_runtime/utils/schemaview.py
+++ b/linkml_runtime/utils/schemaview.py
@@ -122,7 +122,6 @@ class SchemaView(object):
                 namespaces.add_prefixmap(cmap, include_defaults=False)
         return namespaces
 
-
     def load_import(self, imp: str, from_schema: SchemaDefinition = None):
         if from_schema is None:
             from_schema = self.schema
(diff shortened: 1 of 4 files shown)


linkml_runtime/utils/schemaview.py
```diff
@@ -529,6 +531,40 @@ def expand_curie(self, uri: str) -> str:
                     return ns[pfx] + local_id
         return uri
 
+    @lru_cache(CACHE_SIZE)
+    def get_element_by_prefix(
```
cmungall, line 535: perhaps decompose this:

 - a method that takes an actual prefix as input and returns element names, `get_elements_applicable_for_prefix`
 - a method that splits CURIEs (or URIs?) and. extracts the prefix
 - a wrapper that composes these `get_elements_applicable_for_identifier`

sierra-moxon: I went ahead and broke this up into 3 methods and added support (by utilizing an existing function! - thanks for the pointer :) ) for URIs as well.

linkml_runtime/utils/schemaview.py
```diff
@@ -663,6 +666,54 @@ def expand_curie(self, uri: str) -> str:
...
+            The prefix of a CURIE
+
+        Returns
+        -------
+        Optional[str]
+                The model element corresponding to the given prefix as available via
+                the id_prefixes mapped to that element.
+
+        """
+        categories = []
```
cmungall, line 709: minor: categories is a biolinkism - use elements instead? not a big deal as this is not exposed to the caller but still good to use internally consistent variable names

linkml_runtime/utils/schemaview.py
```diff
@@ -663,6 +666,54 @@ def expand_curie(self, uri: str) -> str:
                     return ns[pfx] + local_id
         return uri
 
+    @lru_cache(CACHE_SIZE)
+    def get_elements_applicable_by_identifier(self, identifier: str) -> List[str]:
+        """
+        Get a Model element by prefix.
+
+        Parameters
```
cmungall, line 674: minor: can you change to the rst format used for other methods?

cmungall (approved):

sierra-moxon:
fixed up doc and moved categories -> elements

PR #37, opened 2021-09-09
Method pluralization

sierra-moxon:
I'm happy to go either way on this; probably just a preference thing.


cmungall (approved):

linkml_runtime/utils/schemaview.py
```diff
@@ -1,18 +1,14 @@
 import os
 import uuid
-import logging
```
cmungall, line 3: oops looks like this conflicted with another change

PR #59, opened 2021-10-21
New rdf dumper - bypasses existing jsonld based method and does a direct conversion from elements to rdflib

cmungall:
This is an alternative route to rdf dumping

The existing code (rdf_dumper) uses a strategy of leveraging the generated jsonld_contexts, and using the existing json dumper. However, there are a few things incomplete here (e.g no support for `@embed` in jsonld contexts). @hsolbrig will work on this later.

In the interim, this module uses a direct approach, recursing through the element tree emitting triples. It implement the dumper interface, but it requires passing a SchemaView object


cmungall:
There is a roundtrip test that I believe covers many different cases (inlining, multivalues, enums with meanings, meaningless enums), but more tests welcome!

Question for discussion Friday: should rdflib be the default dumper/loader? with the existing incomplete libs moved to something like jsonld_loader/dumper?

PR #71, opened 2021-11-22
Fix get_mapping_index method 

sierra-moxon:
was interested in using get_mapping_index here, but it seems to return mappings both ways (ie: ks:Activity as an index value and prov:Activity as an index value)?  I would love your feedback here, @cmungall


cmungall (changes requested):
can you address the minor changes suggested in #72 but. otherwise looks good!

cmungall (approved):

PR #72, opened 2021-11-22
monarch notebook + induced slots fix

sierra-moxon:


linkml_runtime/utils/schemaview.py
```diff
@@ -663,6 +690,54 @@ def expand_curie(self, uri: str) -> str:
                     return ns[pfx] + local_id
         return uri
 
+    @lru_cache(CACHE_SIZE)
+    def get_elements_applicable_by_identifier(self, identifier: str) -> List[str]:
+        """
+        Get a Model element by prefix.
+
+        Parameters
```
cmungall, line 698: we are using rst style elsewhere

linkml_runtime/utils/schemaview.py
```diff
@@ -663,6 +690,54 @@ def expand_curie(self, uri: str) -> str:
...
+            The prefix of a CURIE
+
+        Returns
+        -------
+        Optional[str]
+                The model element corresponding to the given prefix as available via
+                the id_prefixes mapped to that element.
+
+        """
+        categories = []
```
cmungall, line 733: this is biolink terminology - can we have a variable name that is more generic, e.g. matching_elements

linkml_runtime/utils/schemaview.py
```diff
@@ -674,22 +749,35 @@ def get_mappings(self, element_name: ElementName = None, imports=True, expand=Fa
...
+            }
+        else:
+            m_dict = {}
         if expand:
             for k, vs in m_dict.items():
                 m_dict[k] = [self.expand_curie(v) for v in vs]
 
         return m_dict
 
+    def get_category_by_mapping(self, mapping_id: URIorCURIE) -> List[str]:
```
cmungall, line 771: how about

`def get_element_by_mapping(self, mapping_vavlue: URIorCURIE)`

linkml_runtime/utils/schemaview.py
```diff
@@ -586,6 +590,30 @@ def slot_leaves(self, imports=True, mixins=True) -> List[SlotDefinitionName]:
                 for c in self.all_slots(imports=imports)
                 if self.slot_children(c, mixins=mixins, imports=imports) == []]
 
+    def is_mixin(self, element_name: Union[ElementName, Element]) -> bool:
+        if isinstance(element_name, Element):
+            return element_name
```
cmungall, line 595: is this right?

linkml_runtime/utils/schemaview.py
```diff
@@ -586,6 +590,30 @@ def slot_leaves(self, imports=True, mixins=True) -> List[SlotDefinitionName]:
...
+        name: str
+            The name or alias of an element
+
+        Returns
+        -------
+        bool
+            true if the element has an inverse
+        """
+        element = self.get_element(name)
+        inverse = element.inverse if isinstance(element, SlotDefinition) else "primary direction"
```
cmungall, line 615: this will return "primary direction" if the element is a class or enum

Maybe restrict the signature to be SlotName or SlotDefinition

In general I would also avoid hardcoding strings...

For slots I think there are 3 cases:

 1. asserted inverse
 2. inferred inverse (e.g. `next(inv.name for inv in view.slots if inv.inverse == s.name)`)
 3. no inverse

The concept of 'primary direction' is biolink specific, although we could push this convention to the linkml level, it may make more sense to have the runtime be more generic and concepts of primary direction be at the application level

linkml_runtime/linkml_model/meta.py
```diff
@@ -1020,7 +1020,12 @@ class SlotDefinition(Definition):
     """
     the definition of a property or a slot
     """
-    _inherited_slots: ClassVar[List[str]] = ["domain", "multivalued", "inherited", "readonly", "ifabsent", "inlined", "inlined_as_list", "key", "identifier", "designates_type", "role", "range", "required", "recommended", "minimum_value", "maximum_value", "pattern", "equals_string", "equals_string_in", "equals_number", "equals_expression", "minimum_cardinality", "maximum_cardinality"]
+    _inherited_slots: ClassVar[List[str]] = ["domain", "multivalued", "inherited", "readonly", "ifabsent", "inlined",
```
cmungall, line 1023: is this a spurious diff from black or edited? Note if edited, it will be overwritten as this file is auto-gen'd from model

PR #80, opened 2021-12-02
add poetry support to the repo

sujaypatil96:
Add poetry specific files: `pyproject.toml` and `poetry.lock` to the repo so that the package can be installed with poetry commands.

Run the tests using poetry to make sure everything still works as expected:

```
poetry run make test
```

This PR closes [issue #465](https://github.com/linkml/linkml/issues/465).


cmungall (changes requested):
- [x] add instructions for developers to CONTRIBUTING.md
      - do we need to do anything in our IDEs?
      - how does this affect pypi releases?
      - who updates lock file, how often?
 - [x] document versioning strategy (this says 1.0.0 which is not correct)
 - [x] remove pipenv stuff
 - [ ] change Makefile (it still assumes pipenv)

> This PR closes issue #465.

not quite - this issue is pan-repo

sujaypatil96:
> add instructions for developers to CONTRIBUTING.md

I've opened an issue for adding poetry specific contribution contribution guidelines to the CONTRIBUTING.md document in issue #83.

> document versioning strategy (this says 1.0.0 which is not correct)

The latest release is tagged with `v1.1.9`. Since we're adding a feature, I think it's fair to bump up the minor version number to  `v1.2.0`?

> remove pipenv stuff, change Makefile

So the Github Actions workflow still uses pipenv, which is why I didn't remove the pipenv specific files. I'll modify the Github Actions workflow to use poetry in a separate PR after this PR has been merged in. See issue

cmungall:
we will stay in 1.1.x for now

ChangeLog
```diff
@@ -3,6 +3,11 @@ CHANGES
 
 * Update to latest model
 
+v1.2.0
```
cmungall, line 6: don't modify ChangeLog as part of this PR. ChangeLog is changed automatically?

sujaypatil96:
Don't commit `poetry.lock` file since `linkml-runtime` is a library. See [docs](https://python-poetry.org/docs/basic-usage/#commit-your-poetrylock-file-to-version-control).

gaurav:
> Don't commit `poetry.lock` file since `linkml-runtime` is a library. See [docs](https://python-poetry.org/docs/basic-usage/#commit-your-poetrylock-file-to-version-control).

Note that elsewhere [the Poetry docs](https://python-poetry.org/docs/libraries/#lock-file) says:
> For your library, you may commit the poetry.lock file if you want to. This can help your team to always test against the same dependency versions. However, this lock file will not have any effect on other projects that depend on it. It only has an effect on the main project.

So it looks to me like there's no downside to committing the lock file for libraries -- they only note that applications that use this library shouldn't expect to use the same version of other dependencies as specified in the lock file. I like having the lockfile around as a record of exactly which packages were used when the library was being developed on -- this can be really useful, especially if the project goes dormant for any reason.

(Arguably, the more useful way of doing this would be for the GitHub Action to log the output of `poetry lock; cat poetry.lock`, so we know both that all tests passed AND which versions of the dependencies were used in order to run it.)

sujaypatil96:
Thank you for the note about the lock file @gaurav. That makes a lot of sense. Yes, you're right in that there's no downside to it. And given that this library will use a simple `pip install linkml-runtime` for installation, I don't see `poetry.lock` getting in the way anywhere either.

sujaypatil96:
CC: @cmungall. At this point, the Github Actions have been updated to use poetry specific commands. Next steps will include removal of unnecessary build tools and files and modifications to the Makefile.

cmungall (approved):

sierra-moxon (approved):
🎉

PR #81, opened 2021-12-03
Linkml issue 497: Inherited class slots are not reflecting the correct range in the python and markdown files

wdduncan:
Fixes https://github.com/linkml/linkml/issues/497

@cmungall The unit test passes. However, if you look a the generated python file `linkml-runtime/tests/test_issues/input/linkml_issue_497.py`, you will see that the [has_input](https://github.com/linkml/linkml-runtime/blob/3e5dc1116be3b4864d4fbe521d4e20c1db8cde53/tests/test_issues/input/linkml_issue_497.py#L96) attribute of `OmicsProcessing` is of type `NamedThing`:

```python
@dataclass
class OmicsProcessing(BiosampleProcessing):
    _inherited_slots: ClassVar[List[str]] = []

    class_class_uri: ClassVar[URIRef] = URIRef("https://example.com/test-inherited-ranges/OmicsProcessing")
    class_class_curie: ClassVar[str] = None
    class_name: ClassVar[str] = "omics processing"
    class_model_uri: ClassVar[URIRef] = URIRef("https://example.com/test-inherited-ranges/OmicsProcessing")

    has_input: Optional[Union[Union[dict, NamedThing], List[Union[dict, NamedThing]]]] = empty_list()
...
```
I did some playing around, and found out that if I remove/comment out the `has input` slot from the `omics processing` class in the yaml file, and include the `slot_usage` like so:
```yaml
    slot_usage:
        has input:
         required: true    
```
the type of the `has_input` attribute in the python file will be `Biosample`

### DIFF
--- a/tests/test_issues/test_linkml_issue_497.py
+++ b/tests/test_issues/test_linkml_issue_497.py
@@ -1,3 +1,5 @@
+import re
+import logging
 from unittest import TestCase, main
 from linkml_runtime.utils.schemaview import SchemaView
 from tests.test_issues.environment import env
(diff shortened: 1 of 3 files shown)


cmungall (changes requested):
I think you meant to put in linkml

however it is still a good schemaview test - you can remove the generated py file and we can keep this test

wdduncan:
@cmungall I added a hacky check for the range of `has input` using a regex on the python file.

tests/test_issues/test_linkml_issue_497.py
```diff
@@ -0,0 +1,34 @@
...
+      has_input_bp = view.induced_slot('has input', 'biosample processing')
+      has_input_op = view.induced_slot('has input', 'omics processing')
+
+      # testing schemaview ranges
+      assert has_input.range == 'named thing'
+      assert has_input_bp.range == 'biosample'
+      assert has_input_op.range == 'biosample'
+
+      # test python file's has_input attribute
+      with open(env.input_path('linkml_issue_497.py')) as f:
```
cmungall, line 26: It looks like you are testing a statically checked in file. If you are intending to test PythonGenerator, then the generators are in the linkml repo.

wdduncan:
Closing this PR, and moving to the LinkML repo.

Issue #76, opened 2021-11-23
Unit tests with Python 3.10 fail


hsolbrig:
When Python 3.10 is included in the github actions with `python-version: [ 3.7.1, 3.8, 3.9, "3.10" ]`, the following message is generated:

```Resolving dependencies...

FAIL

[ResolutionFailure]:   File "/usr/local/lib/python3.8/dist-packages/pipenv/resolver.py", line 743, in _main
[ResolutionFailure]:       resolve_packages(pre, clear, verbose, system, write, requirements_dir, packages, dev)
[ResolutionFailure]:   File "/usr/local/lib/python3.8/dist-packages/pipenv/resolver.py", line 704, in resolve_packages
[ResolutionFailure]:       results, resolver = resolve(
[ResolutionFailure]:   File "/usr/local/lib/python3.8/dist-packages/pipenv/resolver.py", line 685, in resolve
[ResolutionFailure]:       return resolve_deps(
[ResolutionFailure]:   File "/usr/local/lib/python3.8/dist-packages/pipenv/utils.py", line 1370, in resolve_deps
[ResolutionFailure]:       results, hashes, markers_lookup, resolver, skipped = actually_resolve_deps(
[ResolutionFailure]:   File "/usr/local/lib/python3.8/dist-packages/pipenv/utils.py", line 1099, in actually_resolve_deps
[ResolutionFailure]:       resolver.resolve()
[ResolutionFailure]:   File "/usr/local/lib/python3.8/dist-packages/pipenv/utils.py", line 877, in resolve
[ResolutionFailure]:       raise ResolutionFailure(message=str(e))
[pipenv.exceptions.ResolutionFailure]: Warning: Your dependencies could not be resolved. You likely have a mismatch in your sub-dependencies.
  You can use $ pipenv install --skip-lock to bypass this mechanism, then run $ pipenv graph to inspect the situation.
  Hint: try $ pipenv lock --pre if it is a pre-release dependency.
ERROR: Disabling PEP 517 processing is invalid: project specifies a build backend of setuptools.build_meta in pyproject.toml
```

Since this error ONLY occurs with Python 3.10, the issue has to lie in:
a) Maybe the quotes around the 3.10 are causing grief (not likely)?
b) Maybe pipenv version isn't compatible?
c) ???

Looking up the PEP 517 discussions haven't been fruitful. FWIW, the pipenv folks don't seem particularly friendly on this front...

sujaypatil96:
[...]

PR #89, opened 2021-12-10
Fix signature issue in previous issue_76 fix.

hsolbrig:
Also update the publish action to correctly update the
AUTHORS and ChangeLog

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -37,7 +37,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.7.1, 3.8, 3.9]
+        python-version: [3.7.1, 3.8, 3.9, "3.10"]
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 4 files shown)


hsolbrig:
I have no idea what is going on with the python 3.10 in the unit tests.  It works locally, so the only difference I can come up with is that I'm using 3.10.1 and github's latest is still 3.10.0.  Hopefully this will resolve itself in due time?

sujaypatil96:
@hsolbrig: there was increasing advocacy to integrate and use [poetry](https://python-poetry.org/docs/) rather than setuptools with pbr within the group. So the current version of the `main` branch supports build and distribution of the package with poetry.

In the process I also updated the [pypi-publish.yaml](https://github.com/linkml/linkml-runtime/blob/main/.github/workflows/pypi-publish.yaml) workflow to publish the package with poetry commands. setuptools is being deprecated in PR #91. So one of the side effects of setuptools being deprecated is that pbr will be deprecated as well. And since we won't have pbr, we won't have the automated updation of AUTHORS and ChangeLog. Do you have any ideas to circumvent this?

cmungall:
@sujaypatil96 can you make a new PR with the equivalnt changes?

sujaypatil96:
Left a comment on issue #76.

PR #94, opened 2021-12-17
DO NOT MERGE YET-rdflib6

cmungall:


cmungall:
Even though this PR passes, do not merge until 

* https://github.com/linkml/linkml/pull/524

Otherwise linkml will no longer be compatible with linkml-runtime!

PR #95, opened 2021-12-19
fix(prefixes): warn about unrecognized prefix instead of error

noelmcloughlin:
This PR allows documentation to be built despite https://github.com/linkml/linkml-runtime/issues/96

But a solution to #96 would be better of course..

### DIFF
--- a/linkml_runtime/utils/namespaces.py
+++ b/linkml_runtime/utils/namespaces.py
@@ -192,7 +192,9 @@ class Namespaces(CaseInsensitiveDict):
                     return k + ':' + suffix
             return self.join(str(prefix), suffix)
         elif prefix not in self:
-            raise ValueError(f"Unrecognized prefix: {prefix}")
+            # raise ValueError(f"Unrecognized prefix: {prefix}")
+            print(f"WARNING: Unrecognized prefix: {prefix}")
+            return ':' + suffix
         else:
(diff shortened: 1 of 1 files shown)


noelmcloughlin:
image

linkml_runtime/utils/namespaces.py
```diff
@@ -192,7 +192,9 @@ def uri_or_curie_for(self, prefix: Union[str, URIRef], suffix: str) -> str:
                     return k + ':' + suffix
             return self.join(str(prefix), suffix)
         elif prefix not in self:
-            raise ValueError(f"Unrecognized prefix: {prefix}")
+            # raise ValueError(f"Unrecognized prefix: {prefix}")
+            print(f"WARNING: Unrecognized prefix: {prefix}")
```
cmungall, line 196: I would prefer logging.error or something that writes on stderr

cmungall (changes requested):

linkml_runtime/utils/namespaces.py
noelmcloughlin: Yes, this is not ideal solution - I'm happy to close or change the PR while I investigate #96 further to be honest.

noelmcloughlin:
Closing see https://github.com/linkml/linkml-runtime/issues/96

PR #107, opened 2022-01-13
Version support in github actions vs pyproject.toml

sierra-moxon:


linkml_runtime/__init__.py
```diff
@@ -4,6 +4,11 @@
 import rdflib_shim
 shim = rdflib_shim.RDFLIB_SHIM
 
+try:
```
cmungall, line 7: can you add a comment to say why this is necessary?

tests/test_utils/test_version.py
```diff
@@ -0,0 +1,3 @@
+import linkml_runtime
+
+print(linkml_runtime.__version__)
```
cmungall, line 3: can you make it an actual test (and avoid prints)?

cmungall (approved):
let's try it!

tests/test_utils/test_version.py
sujaypatil96: Updated the unit test.

linkml_runtime/__init__.py
sujaypatil96: Added a comment clarifying why we need `importlib-metadata`.

sujaypatil96 (approved):
@sierra-moxon: So it turns out `importlib-metadata` isn't bundled in native Python for versions `<3.8`, so I've added it as a dev dependency in the pyproject.toml file if the version being tested is `<3.8`.

@cmungall: Perhaps we can create a release from the main branch to see how this solution works. We created releases from this branch already and it works like a charm.

cmungall:
go for it!

PR #110, opened 2022-01-21
more details added to CONTRIBUTING.md

sujaypatil96:


cmungall (approved):
I don't think poetry.lock should be updated when making markdown changes - should only change if we explicitly need to bump a version or add a new dependency?

PR #112, opened 2022-01-24
using logging not print

cmungall:


sujaypatil96 (changes requested):
If you look at the official Python logging module [documentation](https://docs.python.org/3/howto/logging.html#advanced-logging-tutorial), you can read about the de facto convention we should follow when creating loggers.

You can do this instead:

```
import logging

logger = logging.getLogger(__name__) # create logger object
# configure logger here if you want to using .basicConfig()

# business logic

logger.warning("")
```

sujaypatil96:
Approving this PR, but open an issue to adopt `logger` objects like above across repos.

sujaypatil96 (approved):

PR #113, opened 2022-01-24
load_source should return lists when lists are passed

cmungall:


hsolbrig:
Is there a unit test?  How do we know whether it worked?

hsolbrig (approved):
Happy to merge, but note that if the code breaks, we'll never know...

cmungall:
Will add tests

Fixes https://github.com/linkml/linkml/issues/589

cmungall:
Tests added!

PR #115, opened 2022-01-31
Test environment update

hsolbrig:


hsolbrig:
Looking at the 3.8 build error -- apparently the dataclasses library changed somewhere IN the python 3.8 branch -- I'm going to have to find the source and find out where the cutoff is.  Hold of on merges until then

PR #131, opened 2022-02-15
attributes take priority if defined for a class

cmungall:


sujaypatil96:
Thank you @cmungall for the fix to the test in `test_inference_utils.py`.

PR #143, opened 2022-03-04
fixes https://github.com/linkml/linkml/issues/634

StroemPhi:
I dont know if this is the way to go, as having this set as default might not be desired for other use cases, but this would resolve my problem reported in https://github.com/linkml/linkml/issues/634.

Still open would be the `json_dumper.dumps()` variant.

### DIFF
--- a/linkml_runtime/dumpers/json_dumper.py
+++ b/linkml_runtime/dumpers/json_dumper.py
@@ -51,7 +51,7 @@ class JSONDumper(Dumper):
             else:
                 return json.JSONDecoder().decode(o)
         return json.dumps(as_json_object(element, contexts, inject_type=inject_type),
-                          default=default,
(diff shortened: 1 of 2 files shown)


StroemPhi:
[5e775a5](https://github.com/linkml/linkml-runtime/pull/143/commits/5e775a5695a4c40d76f35ced6daf822d9f6c92f9) allows for utf-8 characters when using `json_dumper.dumps()`. But again, I'm not sure if this the desired behaviour for all use cases and I'm sorry if I shouldn't have made the JSON dumper edit in this PR.

StroemPhi:
It's probably better to merge #146 instead of this one.

StroemPhi:
fixed better (with tests) in https://github.com/linkml/linkml-runtime/pull/146, so closing this one

PR #146, opened 2022-03-05
More fixes related to unicode (fixes linkml/linkml#634)

dalito:
More fixes related to unicode: (i) in yaml-files and (ii) on windows (linkml/linkml#601).

I also added an "ö" to the test data. That way unicode handling mistakes may show up earlier in the future.

Fixes linkml/linkml#634

### DIFF
--- a/linkml_runtime/dumpers/dumper_root.py
+++ b/linkml_runtime/dumpers/dumper_root.py
@@ -13,7 +13,7 @@ class Dumper(ABC):
         :param to_file: file to dump to
         :@param _: method specific arguments
         """
-        with open(to_file, 'w', encoding="UTF8") as output_file:
+        with open(to_file, 'w') as output_file:
             output_file.write(self.dumps(element, **_))
 
     @abstractmethod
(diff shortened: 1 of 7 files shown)


cmungall (approved):
Thanks, and thanks for the test case too

sujaypatil96 (approved):
Thank you for the test and the fix for [issue #634](https://github.com/linkml/linkml/issues/634) @dalito.

If you check the CONTRIBUTING.md on the repo, you'll see the way we add tests to the existing test suite for specific issues, is by adding minimal tests to the `tests/test_issues` folder rather than modifying the external loader dumper tests. Do you think you would be able to copy over the same content from the test you have here into another test called `test_linkml_issue_634.py` in the `tests/test_issues` folder? You could also copy over the test input file you're using here with the UTF-8 unicode character, or you could create your own minimal schema with just:

```
id: https://w3id.org/linkml/linkml-runtime/test/test_issues/test_linkml_issue_634
name: Joe Schmö
```

Or since we're actually modifying core loader dumper functionality, it's probably okay to leave the test in here too. @cmungall can probably weigh in here and help make a decision?

dalito:
I started with a specific test but then saw that dumping rdf failed also because the Dumper abstract base class in `dumper_root.py` was not using unicode (IIRC). So I thought it is better to address Unicode compatibility in a more general way. But there are also arguments for small isolated test. I don't have a strong preference.

PR #159, opened 2022-03-28
add method to return all_classes, all_slots in alphabetical and/or rank order.

sierra-moxon:
Biolink-Model has a long list of classes.  I these ording methods in my jinga templates to return classes in alphabetical order (or rank order) for rendering.

### DIFF
--- a/linkml_runtime/utils/schemaview.py
+++ b/linkml_runtime/utils/schemaview.py
@@ -200,6 +200,7 @@ class SchemaView(object):
         """
         return self._get_dict(CLASSES, imports)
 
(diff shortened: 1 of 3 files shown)


cmungall (commented):
there may be different orderings for different use cases:

1. the order they are listed in the yaml (this is preserved and may be meaningful)
2. lexical order (this)
3. order determined by explicit specification of `rank`
4. pre or post order traversal over parent graph
5. ...

maybe these should be a parameter that controls this?

sierra-moxon:
talked to Chris - this is ready to go if tests pass.

PR #160, opened 2022-03-29
remove pyyaml version pin

sujaypatil96:
Change pyproject.toml and poetry.lock to not pin the version of pyyaml. We already modified the code to accommodate fixes to `yaml.load()`.


sujaypatil96:
Work on analogous PR in linkml core that unpins pyyaml.

cmungall (approved):

PR #162, opened 2022-04-09
Enhance import_map functionality to support namespace as well as indi…

hsolbrig:
…vidual file mapping.

Also separated out import mapping code so that linkml can use the same code as the utilities.

### DIFF
--- a/linkml_runtime/utils/context_utils.py
+++ b/linkml_runtime/utils/context_utils.py
@@ -50,7 +50,7 @@ def merge_contexts(contexts: CONTEXTS_PARAM_TYPE = None, base: Optional[Any] = N
         JsonObj(**{"@context": context_list[0] if len(context_list) == 1 else context_list})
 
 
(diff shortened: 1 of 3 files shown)


hsolbrig:
@cmungall  - I need this to be in pypi before I can commit a corresponding linkml change.  What is the process these days?

cmungall:
same as before - just make a new github release, actions will propagate to pypi

See the docs: https://github.com/linkml/linkml-runtime/blob/main/CONTRIBUTING.md#release-process

linkml_runtime/utils/context_utils.py
```diff
@@ -50,6 +50,17 @@ def to_file_uri(fname: str) -> str:
         JsonObj(**{"@context": context_list[0] if len(context_list) == 1 else context_list})
 
 
+def map_import(importmap: Dict[str, str], namespaces: Callable[[None], "Namespaces"], imp: Any) -> str:
+    sname = str(imp)
```
cmungall, line 54: a docstring would be great here

cmungall (approved):

Issue #163, opened 2022-04-15
ImportMap strips trailing separators on local files


hsolbrig:
An import map of the form:
```json
{  "base:": "base/"}
```
Maps `base:abc` to `<path>/baseabc` instead of `path/base/abc`.  This is because `os.path.abspath` strips trailing separators.

sierra-moxon:
closing as completed.

PR #164, opened 2022-04-15
Fix issue #163

hsolbrig:
Still not sure this works correctly in a Windows environment.  We really need
to get windows added to our test suite

### DIFF
--- a/linkml_runtime/utils/context_utils.py
+++ b/linkml_runtime/utils/context_utils.py
@@ -90,7 +90,7 @@ def parse_import_map(map_: Optional[Union[str, Dict[str, str], TextIOWrapper]],
             if ':' not in v:
                 v = os.path.join(base, v)
                 if ':/' not in v:
-                    v = os.path.abspath(v)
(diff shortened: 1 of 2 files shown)


cmungall:
@dalito has been kind enough to contribute PRs to fix windows issues - he may have some ideas for strategies

dalito:
This fix fails on windows, see windows CI run added in #165

linkml_runtime/utils/context_utils.py
```diff
@@ -90,7 +90,7 @@ def parse_import_map(map_: Optional[Union[str, Dict[str, str], TextIOWrapper]],
             if ':' not in v:
                 v = os.path.join(base, v)
                 if ':/' not in v:
-                    v = os.path.abspath(v)
+                    v = os.path.abspath(v) + (os.sep if v.endswith(os.sep) else '')
             outmap[k] = v
```
dalito, line 94: Could you just write: (this passes the tests also on windows)
```
            if ':' not in v:
                v = os.path.join(os.path.abspath(base), v)
            outmap[k] = v
```

PR #165, opened 2022-04-17
Add windows to CI test-matrix

dalito:
A first try. I have not yet conributed chanes to CI workflows, so not sure how/if it is picked up from the PR.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -8,11 +8,18 @@ on: [pull_request]
 
 jobs:
   test:
-
-    runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


cmungall:
Well it seems to be picked up ok - apologies that the test output is hard to read - does this give you the info you need?

dalito:
Here is a slightly different fix for #163. This one passes the tests also on windows.

Sorry, my git-client accidentally included two ci-configs that I had to remove again in the next commit.

sujaypatil96 (approved):
Harold's test case from PR #164 seems to be going through on all the windows enabled Action runs so everything seems to be in order for a merge.

hsolbrig (commented):
Say, thanks for adding the windows test.  Fix works, we're good to go!

PR #168, opened 2022-04-28
Proposed fix for issue #718

hsolbrig:
The `dataclasses` package uses the built-in class signature detection to throw a
TypeError if an unrecognized parameter is passed in a constructor.  This:
a) makes it difficult or impossible to print information about where the parameter came from in a yaml (other file)
b) limits our options as to whether we go ahead and continue processing.

The current fix (prior to this submission) was overriding the `_init_fn` in the
dataclasses library, which turned out to be anything but stable.  This new fix,
instead, tries to focus on the `_create_fn`, whose signature and behavior is
considerably more stable.

IF this fix works, we want to proceed by changing the python generators to emit something like:
```python
import linkml_runtime.utils.dataclass_extensions_376 as dc_tweak
# This line makes certain that the import remains
DC_IN = dc_tweak.DC_CREATE_FN
```
And, finally, renaming `dataclass_extensions_376` to something more reasonable.


cmungall:
https://github.com/linkml/linkml/pull/718

PR #169, opened 2022-04-29
Include a version of the python 3.10 dataclass patch

hsolbrig:
See: https://github.com/python/cpython/issues/92052 for
details. This patch should add the patch to any module that imports linkml_runtime.

This PR is not yet ready for prime time -- we will be testing it with the linkml package first and using it to check options
for the [pydantic](https://pydantic-docs.helpmanual.io/) wrapper


cmungall:
@hsolbrig what are the next steps here?

PR #170, opened 2022-05-06
Updated linkml-model to 1.2.2

cmungall:


turbomam:
Approved by @cmungall in Slack DM

PR #171, opened 2022-05-09
turn off info logging for illegal prefix while we clean up these

sierra-moxon:
these three prefixes from identifiers.org generate a lot of output in modularized schemas (tend to obfuscate other warnings/errors).  would it be ok to turn this off for a while until we can fix these three prefixes? 
INFO:Namespaces:biocontext map idot_context has illegal prefix: 2D-PAGE.PROTEIN
INFO:Namespaces:biocontext map idot_context has illegal prefix: 3DMET
INFO:Namespaces:biocontext map idot_context has illegal prefix: MMMP:BIOMAPS


cmungall:
I don't understand why logging happens different in this python class than elsewhere

PR #181, opened 2022-06-03
WIP: Fixed RDF prefix issue in rdflib dumper

victoriasoesanto:


Pipfile
```diff
@@ -0,0 +1,12 @@
+[[source]]
```
cmungall, line 1: I don't think you intended to add the Pipfile?

It's a good idea to check what files you are committing with `git status` in advance - only files you have intentionally modified should be present

.gitignore
```diff
@@ -140,3 +140,7 @@ Pipfile.lock
 
 # No Pycharm
 .idea/
+.DS_store
```
cmungall, line 143: this is harmless but did you mean to include this?

It's best to treat each PR as a separate concern - if you want to modify the .gitignore do it as a separate PR (but this is OK for now)

tests/test_loaders_dumpers/test_rdf_prefixes.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+class RdfLibDumperTestCase(unittest.TestCase):
+
+    def test_rdflib_dumper(self):
+        view = SchemaView(SCHEMA)
+        container = yaml_loader.load(DATA, target_class=Container)
+        rdflib_dumper.dump(container, schemaview=view, to_file=OUT)
+        g = Graph()
+        g.parse(OUT, format='ttl')
```
cmungall, line 22: every test should contain at least one assertion, so that it's actually a test

This is a difficult one to test, as the prefix declarations disappear on parsing with rdflib

I would test by doing a direct read using `open(..)` on `OUT`, and simply scan the file to see if it includes the intended header. This might be somewhat rigid but I think we can expect the rdflib turtle output to be reasonably stable

in fact just checking for the string 'prefix <expectedPrefix>:' should befine

cmungall (changes requested):
This is great, thanks so much! Can you address the comments then we can review. Also, remember to always link your PRs to an issue

.gitignore
sierra-moxon: @victoriasoesanto - please don't delete the .gitignore file from this PR.  deleting the Pipfile will be good, but not the .gitignore.

PR #184, opened 2022-06-08
WIP: RDF prefix issue fix in rdflib dumper

victoriasoesanto:
Fixes https://github.com/linkml/linkml/issues/429

### DIFF
--- a/tests/test_loaders_dumpers/test_issue_429.py
+++ b/tests/test_loaders_dumpers/test_issue_429.py
@@ -1,0 +1,47 @@
+import os
+import unittest
+from rdflib import Graph, Literal
+from rdflib import Namespace
+from rdflib.namespace import RDF
+from linkml_runtime.dumpers import rdflib_dumper
+from linkml_runtime.loaders import yaml_loader
+from linkml_runtime.utils.schemaview import SchemaView
+from tests.test_loaders_dumpers import INPUT_DIR, OUTPUT_DIR
+from tests.test_loaders_dumpers.models.personinfo_test_issue_429 import Container
+
+SCHEMA = os.path.join(INPUT_DIR, 'personinfo_test_issue_429.yaml')
+DATA = os.path.join(INPUT_DIR, 'example_personinfo_test_issue_429_data.yaml')
+OUT = os.path.join(OUTPUT_DIR, 'example_personinfo_test_issue_429_data.ttl')
+
+ORCID = Namespace('https://orcid.org/')
+personinfo = Namespace('https://w3id.org/linkml/examples/personinfo/')
+SDO = Namespace('http://schema.org/')
+
+
+class RdfLibDumperTestCase(unittest.TestCase):
+
+    def test_rdflib_dumper(self):
+        view = SchemaView(SCHEMA)
+        container = yaml_loader.load(DATA, target_class=Container)
+        rdflib_dumper.dump(container, schemaview=view, to_file=OUT)
+        g = Graph()
+        g.parse(OUT, format='ttl')
+        self.assertIn((ORCID['1234'], RDF.type, SDO.Person), g)
+        self.assertIn((ORCID['1234'], personinfo.full_name, Literal("Clark Kent")), g)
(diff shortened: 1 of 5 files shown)


tests/test_loaders_dumpers/test_issue_429.py
```diff
@@ -0,0 +1,47 @@
...
+        self.assertIn((ORCID['1234'], personinfo.full_name, Literal("Clark Kent")), g)
+        self.assertIn((ORCID['1234'], personinfo.age, Literal("32")), g)
+        self.assertIn((ORCID['1234'], personinfo.phone, Literal("555-555-5555")), g)
+        self.assertIn((ORCID['4567'], RDF.type, SDO.Person), g)
+        self.assertIn((ORCID['4567'], personinfo.full_name, Literal("Lois Lane")), g)
+        self.assertIn((ORCID['4567'], personinfo.age, Literal("33")), g)
+        self.assertIn((ORCID['4567'], personinfo.phone, Literal("555-555-5555")), g)
+
+    def test_prefixes(self):
+        with open(OUT, encoding='UTF-8') as file:
```
cmungall, line 39: Note that this introduces dependencies between tests. It assumes the tests will be executed by the test framework in order, and that the developer who it iteratively testing will run things in order.

I would just move this into test_rdflib_dumper

tests/test_loaders_dumpers/test_issue_429.py
```diff
@@ -0,0 +1,47 @@
...
+            for line in file:
+                if 'prefix' in line:
+                    print(line)
```
cmungall, line 42: we should avoid prints in tests, as this makes the test output very verbose and hard to debug

you can use logging.info

But more importantly, this test doesn't test for anything! Recall that this PR is to ensure certain behavior at the turtle syntactic level. I think you want to be testing:

1. that orcid and sdo prefixes are in the prefix with the expected short form. E.g. check for strings `prefix: ORCID` and `prefix: SDO`
2. that CURIEs are used in the main body of the turtle. E.g. check for strings `sdo:Person` and `ORCID:1234`

tests/test_loaders_dumpers/test_issue_429.py
```diff
@@ -0,0 +1,47 @@
...
+SCHEMA = os.path.join(INPUT_DIR, 'personinfo_test_issue_429.yaml')
+DATA = os.path.join(INPUT_DIR, 'example_personinfo_test_issue_429_data.yaml')
+OUT = os.path.join(OUTPUT_DIR, 'example_personinfo_test_issue_429_data.ttl')
+
+ORCID = Namespace('https://orcid.org/')
+personinfo = Namespace('https://w3id.org/linkml/examples/personinfo/')
+SDO = Namespace('http://schema.org/')
+
+
+class RdfLibDumperTestCase(unittest.TestCase):
```
cmungall, line 21: it's good to name the class and tests to something specific to what is being tested (we need to update our guidelines on this)

cmungall (changes requested):

tests/test_loaders_dumpers/test_issue_429.py
```diff
@@ -0,0 +1,54 @@
...
+class RdfLibPrefixTestCase(unittest.TestCase):
+    SCHEMA = os.path.join(INPUT_DIR, 'personinfo_test_issue_429.yaml')
+    DATA = os.path.join(INPUT_DIR, 'example_personinfo_test_issue_429_data.yaml')
+    OUT = os.path.join(OUTPUT_DIR, 'example_personinfo_test_issue_429_data.ttl')
+
+    ORCID = Namespace('https://orcid.org/')
+    personinfo = Namespace('https://w3id.org/linkml/examples/personinfo/')
+    SDO = Namespace('http://schema.org/')
+
+    def __init__(self, *args, **kwargs):
```
cmungall, line 22: you might want to consider setUp() here, see https://docs.python.org/3/library/unittest.html#organizing-test-code

cmungall (approved):

PR #185, opened 2022-06-08
add a simple method to return multivalued slots

sierra-moxon:
better might be to have a method that returns all "x" slots where "x" is any metadata component of a slot (e.g.: all "identifier" slots, all "reflexive" slots, etc)?

@turbomam @hrshdhgd - is this what you are looking for?


cmungall (changes requested):
1. I think this only gets slots *asserted* to be multivalued - multivalued is an inherited slot, so it could be declared on a parent (we do this in the metamodel a bit)
2. the SV class could get unwieldy if there are too many predefined operations. I think a few are OK, but I would rather have more composable methods

How about a method `is_multivalued(self, slot, class=None)` that does any necessary inference

behind the scenes it just composes class_induced_slot (for which the class argument is now optional) with a test of the (inferred) slot for `.multivalued`

not quite as convenient as a method that returns a list, but I think all users of SV should be familiar with list comprehensions and able to write code like

`[s for s in sv.all_slots() if sv.is_multivalued(s)]`

(but we could probably do with more documentation/tutorials here!)

linkml_runtime/utils/schemaview.py
```diff
@@ -738,11 +738,21 @@ def is_multivalued(self, slot_name: SlotDefinition) -> bool:
...
         returns True if slot is multivalued, else returns False
         :param slot_name: slot to test for multivalued
-        :return boolean: 
+        :return boolean:
         """
         induced_slot = self.induced_slot(slot_name)
         return True if induced_slot.multivalued else False
 
+    @lru_cache()
+    def slot_has_quality(self, slot_name: SlotDefinition, quality_to_check) -> bool:
```
cmungall, line 747: missing type annotation and param in docstring?

Not sure about naming. This is hard without getting all meta, but perhaps unavoidable. I'd rather avoid introducing new terms like "quality"

- slot_has_metaslot
- slot_has_slot_assignment (we use "slot assignment" in the spec)

I also recommend addition an optional class name argument - if this is present then class induced slot is used

cmungall: ah I see this is intended for boolean only - maybe have more defensive checking

or is the intent this for both 

1. slots with range boolean
2. slots that are not required, checking for whether the optional value is set
3. checking for empty strings and integers != 0

That feels like overloading. I would stick to 1

for checking for booleans, perhaps for a name

- slot_is_true_for?
- slot_has_property
- ...?

cmungall (changes requested):

linkml_runtime/utils/schemaview.py
sierra-moxon: with the changes - do you still think I should handle class_name argument?

cmungall (approved):

PR #187, opened 2022-06-17
unit test for linkml issue 840

sujaypatil96:
* Test case for https://github.com/linkml/linkml/issues/840
* Quick type specification bug fix on all_schema()


sierra-moxon:
@sujaypatil96 - is this PR open to wait for a fix to the original issue?

PR #195, opened 2022-07-20
uncamelcase method 

sierra-moxon:
working on replacing functionality from biolink-model-toolkit.  Translator biolink model components come back from queries of their architecture in camelcase already (SmallMolecule or biolink:SmallMolecule vs. small molecule).  This is just a helper method to convert back to model-case to use SchemaView to retrieve classes instead of biolink-model-toolkit.  Is it overkill to think about making sv.get_class() etc. recognize any form of class name (I can work with this method alone for my needs, I could also put this method in my own code)?

### DIFF
--- a/linkml_runtime/utils/formatutils.py
+++ b/linkml_runtime/utils/formatutils.py
@@ -9,6 +9,16 @@ ws_pattern = re.compile(r'\s+')
 us_pattern = re.compile(r'_+')
 
 
+def uncamelcase(txt: str) -> str:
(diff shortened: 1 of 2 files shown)


cmungall (approved):
Thanks!

After merging let's make sure we unify methods - add_titles should probably use your code here https://github.com/linkml/linkml/blob/6775314cd165fab140b7d4daade8aa85310ee2c4/linkml/utils/schema_fixer.py#L40

PR #197, opened 2022-08-05
implement method to get classes associated with slot

sujaypatil96:
Implement a method to get all classes that use a given slot.

### DIFF
--- a/linkml_runtime/utils/schemaview.py
+++ b/linkml_runtime/utils/schemaview.py
@@ -1,4 +1,5 @@
 import os
+from unicodedata import name
 import uuid
 import logging
 import collections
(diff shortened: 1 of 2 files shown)


linkml_runtime/utils/schemaview.py
```diff
@@ -1,4 +1,5 @@
 import os
+from unicodedata import name
```
sujaypatil96, line 2: Remove.

turbomam:
A slot can be used legally within a class if
- the class mentions is in it's `slots`
- the class mentions it in it's `attributes`
- the class mentions it in `slot_usage` without being mentioned in `slots` (which is allowed but warned against)
- the class has a ancestor class that uses the slot

we could create a single method with flags for each case, returning a single unqualified list.

if any of those conditions are difficult to process, that case could be omitted for now (good idea, @kevinschaper)

sujaypatil96:
Thank you for breaking up the various possible use cases in which we can determine the classes associated with a slot.

* when `direct` is set to True, we are computing the first use case, i.e., slots mentioned in `slots` property of a class
* when `induced` is True, we are computing the second and third use cases together, slots mentioned under `attributes` and `slot_usage` properties on a class

turbomam:
> @turbomam: thank you for breaking up the various possible use cases in which we can determine the classes associated with a slot.
> 
> * when `direct` is set to True, we are computing the first use case, i.e., slots mentioned in `slots` property of a class
> * when `induced` is True, we are computing the second and third use cases together, slots mentioned under `attributes` and `slot_usage` properties on a class

OK, does that mean that slots asserted by ancestors won't appear in this first solution? We should come up with some test cases.

linkml_runtime/utils/schemaview.py
```diff
@@ -1292,6 +1293,35 @@ def slot_range_as_union(self, slot: SlotDefinition) -> List[EnumDefinitionName]:
...
+        :param direct: list all slots directly in the domain of a class, defaults to True
+        :param induced: list all induced slots in the domain of a class, defaults to False
+        :return: list of slots based on above parameters
+        """
+        slots_list = [] # list of all direct or induced slots
+
+        for c_name, c in self.all_classes().items():
+            
+            # check is slot is direct specification on class
+            if direct == True:
```
cmungall, line 1309: replace with `if direct:`

linkml_runtime/utils/schemaview.py
```diff
@@ -1292,6 +1293,35 @@ def slot_range_as_union(self, slot: SlotDefinition) -> List[EnumDefinitionName]:
                 range_union_of.append(x.range)
         return range_union_of
 
+    def get_classes_by_slot(self, slot: SlotDefinition, direct: bool, induced: bool) -> List[ClassDefinitionName]:
+        """Get all classes that use a given slot, either as a direct or induced slot.
+
+        :param slot: slot in consideration
+        :param direct: list all slots directly in the domain of a class, defaults to True
```
cmungall, line 1300: consider if you really need two boolean parameters here. If both are specified, you raise an error (good). But if neither are specified (which is the default) then this returns an empty list.

if you look at other methods you will see it's more idiomatic to have the method always return direct and to have a single boolean param to determine whether this should be supplemented with induced

cmungall (changes requested):
thanks!

note there is a method for getting usages which is a superset of this, but it's useful to have something like this

cmungall (approved):

PR #199, opened 2022-09-01
add schemaview method to compute class siblings

sujaypatil96:
Method to compute all the siblings of a given input class.

### DIFF
--- a/linkml_runtime/utils/schemaview.py
+++ b/linkml_runtime/utils/schemaview.py
@@ -1,4 +1,5 @@
 import os
+from ssl import DefaultVerifyPaths
 import uuid
 import logging
 import collections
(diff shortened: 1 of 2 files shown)


cmungall (changes requested):
What's the use case for this?

This seems a little overfitted towards the assumptions of structure in one schema

Consider a schema where there is a root mixin. that it broadly used by multiple classes. These classes will all be "singlings"

Did you mean to include parameters in the sgnature to control behavior for mixins

In general for this class and any other class it's useful to look at precedent for similar methods, and the signatures of methods you are calling - ask if the default parameters are always what the user wants, if not allow pass through

sujaypatil96:
The use case is that we wanted to see, for example, [Biosample](https://microbiomedata.github.io/nmdc-schema/Biosample/) class in the NMDC Schema has the following siblings: `DataObject`, `GeneProduct`, `Instrument`, `OntologyClass`, `Person`, `Study`.

Hm, I hadn't considered the handling of root mixins.

cmungall:
> And show the relationships between Biosample and each of the above listed classes as https://github.com/linkml/linkml-runtime/pull/199#issuecomment-1234907887 in the mermaid diagram.

We definitely want this - but this has nothing to do with siblings

PR #204, opened 2022-10-04
Materialize structured patterns on class slot usages and attributes

### LINKED ISSUE #203: `SchemaView.materialize_patterns()` does not account for `slot_usage`
If I have a schema that contains the following:

```yaml
# test.yaml
settings:
  generated_id: "[a-z0-9]{6,}"

classes:
  MyClass:
    slots:
      - id

slots:
  id:
    structured_pattern:
      syntax: "prefix:{generated_id}"
```

And I call the `materialize_patterns` method on a `SchemaView` instance for that schema, I see that the `pattern` meta-slot is populated as expected:

```python
>>> from linkml_runtime import SchemaView
>>> view = SchemaView('test.yaml')
>>> view.schema.slots['id'].pattern is None
True
>>> view.materialize_patterns()
>>> view.schema.slots['id'].pattern
'prefix:[a-z0-9]{6,}'
```

However if the `structured_pattern` is instead set on the `slot_usage` of a class:

```yaml
# test2.yaml
settings:
  generated_id: "[a-z0-9]{6,}"

classes:
  MyClass:
    slots:
      -id
  FancyIdClass:
    is_a: MyClass
    slot_usage:
      id:
        structured_pattern:
          syntax: "prefix:{generated_id}"

slots:
  id:
```

Calling `materialize_patterns` on the `SchemaView` instance doesn't seem to have any effect. I suppose I would expect the `pattern` to be materialized on the `slot_usage` definition instead in this case.

```python
>>> view = SchemaView('test2.yaml')
>>> view.schema.classes['FancyIdClass'].slot_usage['id'].pattern is None
True
>>> view.materialize_patterns()
>>> # I think I would expect this to not be None at this point
>>> view.schema.classes['FancyIdClass'].slot_usage['id'].pattern is None
True
```

pkalita-lbl:
These changes extend the functionality of `SchemaView.materalize_patterns` to also include class slot usage and attributes. As a side effect it also slightly changes how `structured_pattern`s are resolved.

Previously there was a utility method called `generate_patterns` which iterated over all of a schema's slots to precompute all of the resolved patterns. Then `materialize_patterns` would iterate over the schema's slots again and populate the schema with the results from `generate_patterns`. 

With these changes, `materialize_patterns` iterates over slots and classes to account for `slot_usage` and `attributes`. In order to not duplicate logic, `generate_patterns` has been replaced by a class called `PatternResolver` which expands `structured_pattern`s on the fly, caching results as it goes. The central logic of how the expansion is done is unchanged.

Fixes #203

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -140,3 +140,4 @@ Pipfile.lock
 
 # No Pycharm
 .idea/
+.vscode
--- a/linkml_runtime/utils/pattern.py
+++ b/linkml_runtime/utils/pattern.py
@@ -2,6 +2,30 @@ from functools import lru_cache
 import re
 from typing import Dict
 
+# We might want to deprecate this method in favor of PatternResolver in the future
+def generate_patterns(schema_view) -> Dict[str, str]:
+    """Generates a dictionary of slot patterns corresponding to
+    the structured patterns in the settings.
+    :param schema_view: SchemaView object with LinkML YAML
+        already loaded
+    :return generated_patterns: dictionary with the 
+        expanded structured patterns 
+    """
+
+    resolver = PatternResolver(schema_view)
+
+    # dictionary with structured patterns in the key and
+    # expanded, or materialized patterns as values
+    generated_patterns = {}
+
+    for _, slot_defn in schema_view.all_slots().items():
+        if slot_defn.structured_pattern:
+            struct_pat = slot_defn.structured_pattern
+            pattern = struct_pat.syntax
+            generated_patterns[pattern] = resolver.resolve(pattern)
+
+    return generated_patterns
+
 
 class PatternResolver():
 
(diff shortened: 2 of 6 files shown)


cmungall (commented):
What about:

https://github.com/linkml/linkml/blob/bb6edd28be3eed19a5684107b87a60d9d035aad7/linkml/generators/linkmlgen.py#L12

To deploy this you will need to first pin the main linkml repo

alternatively why not retain a stub generate_patterns that wraps the OO method, we can deprecate later

pkalita-lbl:
That's a good catch. I didn't think that `generate_patterns` would have been used outside of this project. Although, oddly that's an unused import in the `linkml` project. Still, you make a valid point that we shouldn't just yank away an ostensibly public function without warning. I will re-create a `generate_patterns` method that delegates to `PatternResolver`.

sujaypatil96 (approved):

cmungall (approved):

PR #205, opened 2022-10-10
csv testing and bug fixes

joshmoore:
This primarily adds the test cases mentioned under https://github.com/linkml/linkml/issues/947. In debugging, there were a number of locations where an explicit `None` check helped with determining what was wrong which I've included. Please let me know if those should be separated out.

Note: this is currently failing both for all attributes as well as all slots with:

```
>           self.objectB = Object(**as_dict(self.objectB))
E           TypeError: tests.test_loaders_dumpers.models.table.Object() argument after ** must be a mapping, not str
```

### DIFF
--- a/linkml_runtime/loaders/rdflib_loader.py
+++ b/linkml_runtime/loaders/rdflib_loader.py
@@ -198,6 +198,8 @@ class RDFLibLoader(Loader):
     def _get_id_dict(self, node: VALID_SUBJECT, schemaview: SchemaView, cn: ClassDefinitionName) -> ANYDICT:
         id_slot = schemaview.get_identifier_slot(cn)
         if not isinstance(node, BNode):
+            if id_slot is None:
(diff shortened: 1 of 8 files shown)


sierra-moxon:
Hi @joshmoore  - sorry this took so long to review; I'd like to take a look and get it resolved.  Unfortunately, I don't have permission here to resolve the conflicts.  (thank you for giving maintainers access to the PR - I am not sure why I can't resolve the conflicts with that permission but the button is unavailable to me).  Can you resolve the conflicts here?  Then I can walk through and get the tests passing.  Thanks in advance!

joshmoore:
@sierra-moxon, no worries (though odd): done.

nlharris:
It looks like this has gone stale while waiting to be reviewed

joshmoore:
Please let me know if there's something needed from my side.

PR #213, opened 2022-10-13
enum permissible value ancestors and parents methods

sierra-moxon:
Adding methods to navigate "is_a" hierarchies for permissible values in enumerations.


linkml_runtime/utils/schemaview.py
```diff
@@ -51,11 +53,14 @@ def _closure(f, x, reflexive=True, depth_first=True, **kwargs):
...
         visited.append(i)
-        vals = f(i)
-        for v in vals:
-            if v not in visited:
-                todo.append(v)
-                if v not in rv:
+        if kwargs.get('enum', False):
+            vals = f(i, kwargs.get('enum'))
+        else:
+            vals = f(i)
```
pkalita-lbl, line 59: If this is necessary I would suggest just unconditionally passing all all keyword arguments to `f` instead of adding enum-specific stuff to this generic function:

```python
vals = f(i, **kwargs)
```

**But** I don't think that's really necessary anyway. See my later comment on `permissible_value_ancestors`.

tests/test_utils/test_schemaview.py
```diff
@@ -22,6 +22,19 @@
 
 class SchemaViewTestCase(unittest.TestCase):
 
+    def test_schemaview_enums(self):
+        view = SchemaView(SCHEMA_NO_IMPORTS)
+        for en, e in view.all_enums().items():
+            if e.name == "Animals":
+                for pv, v in e.permissible_values.items():
+                    if pv == "CAT":
+                        assert view.permissible_value_ancestors(pv, e.name) == ['CAT']
```
pkalita-lbl, line 31: Consider using `TestCase` assertion methods instead of bare `assert`s! They provide much better diagnostic messages in the event of a failure. https://docs.python.org/3/library/unittest.html#assert-methods

linkml_runtime/utils/schemaview.py
sierra-moxon: fixed.

tests/test_utils/test_schemaview.py
sierra-moxon: alright, I can clean up the inconsistent use for this class.  It looks like we have a mix of both styles.

pkalita-lbl (approved):

PR #223, opened 2022-11-16
Allow for non-CURIE identifiers to be encoded.

cmungall:
Fixes https://github.com/linkml/linkml/issues/576

This allows for controlling the bahvior of URI encoding
when a string identifier slot is used. Addition of a percent_encoded
annotation forces encoding plus the base URI


joshmoore:
@cmungall: aware of @hsolbrig's comments in https://github.com/linkml/linkml/issues/576, do you see this as being ready to be adopted or should I hold off for it to show up in the public docs?

PR #227, opened 2022-11-18
Support Pydantic BaseModel in loaders and dumpers

kevinschaper:


kevinschaper:
I see `TypeError: Subscripted generics cannot be used with class and instance checks` so it looks like I was overly optimistic with: 
```
        if isinstance(results, Union[BaseModel, YAMLRoot]):
            return results
        else:
            raise ValueError(f'Result is not an instance of BaseModel or YAMLRoot: {type(results)}')
```

and I need to change that to to an OR rather than a union

kevinschaper:
I'm a little surprised that this is passing with only changes to the method signatures, it sounded like we thought we'd explicitly handle BaseModel differently than YAMLRoot. It could be that we need some tests that try to use BaseModel?

cmungall:
looks good so far! Next would be to implement the switch

kevinschaper:
Here's where I'm muddling through @hsolbrig - aside from breaking all of the tests, how do these changes look so far?

(I regret changing all of the method signatures first, since it does add a lot of noise to the PR)

kevinschaper:
currently failing with: 

```
File "pydantic/main.py", line 341, in pydantic.main.BaseModel.__init__
[246](https://github.com/linkml/linkml-runtime/actions/runs/4599915548/jobs/8125893125?pr=227#step:6:247)
pydantic.error_wrappers.ValidationError: 1 validation error for Dataset
[247](https://github.com/linkml/linkml-runtime/actions/runs/4599915548/jobs/8125893125?pr=227#step:6:248)
code_systems -> id
[248](https://github.com/linkml/linkml-runtime/actions/runs/4599915548/jobs/8125893125?pr=227#step:6:249)
  value is not a valid dict (type=type_error.dict)
```

My memory of this is a little hazy, but I think this is why I created https://github.com/linkml/linkml/issues/1304 - but really, rather than give that job to pydanticgen, I need to normalize before sending to the pydantic class

kevinschaper:
I'm going to look at moving this forward by using the normalize kitchen sink data file (and maybe corresponding schema?) that pydanticgen is using in the main linkml repo

cmungall:
I think the data and schema should be in canonical form, but not expanded. E.g. leave prefixes as key-value pairs.

linkml_runtime/dumpers/csv_dumper.py
```diff
@@ -15,7 +16,7 @@
 
 class CSVDumper(Dumper):
 
-    def dumps(self, element: YAMLRoot,
+    def dumps(self, element: Union[BaseModel, YAMLRoot],
```
kevinschaper, line 19: There are two basic changes that happened in this PR, this is the first, where the YAMLRoot type was used, I changed to a union of BaseModel and YAMLRoot

linkml_runtime/dumpers/rdf_dumper.py
```diff
@@ -77,15 +78,19 @@ def dump(self, element: YAMLRoot, to_file: str, contexts: CONTEXTS_PARAM_TYPE =
...
+    def dumps(self, element: Union[BaseModel, YAMLRoot], contexts: CONTEXTS_PARAM_TYPE = None, fmt: Optional[str] = 'turtle') -> str:
         """
         Convert element into an RDF graph guided by the context(s) in contexts
         :param element: element to represent in RDF
         :param contexts: JSON-LD context(s) in the form of a file or URL, a json string or a json obj
         :param fmt: rdf format
         :return: rdflib Graph containing element
         """
+        if isinstance(element, BaseModel):
+            element = element.dict()
```
kevinschaper, line 94: This is the second part of the main theme of this PR, use pydantic's `.dict()` method on anything that is an instance of `BaseModel` and from there the code can carry on doing whatever it was doing and not need to care whether the instance was a BaseModel or a YAMLRoot

linkml_runtime/dumpers/yaml_dumper.py
```diff
@@ -1,18 +1,20 @@
...
 class YAMLDumper(Dumper):
 
-    def dumps(self, element: YAMLRoot, **kwargs) -> str:
+    def dumps(self, element: Union[BaseModel, YAMLRoot], **kwargs) -> str:
         """ Return element formatted as a YAML string """
         # Internal note: remove_empty_items will also convert Decimals to int/float;
         # this is necessary until https://github.com/yaml/pyyaml/pull/372 is merged
-        return yaml.dump(remove_empty_items(element, hide_protected_keys=True),
+
+        dumper_safe_element = element.dict() if isinstance(element, BaseModel) else element
```
kevinschaper, line 16: Look at that fancy ternary situation, but same deal, element.dict() on a pydantic should make it behave like a YAMLRoot for dumping

cmungall: not totally sure wym about ternary... but this has the side effect of making YAMLRoot dumping more verbose?

kevinschaper: Good thing to test either way though!

cmungall (approved):

PR #230, opened 2022-11-29
include subsets in merge_schemas method.

cmungall:
Fixes https://github.com/linkml/linkml/issues/1143

Also fixes problem whereby the merge order of enums is not respected

### DIFF
--- a/linkml_runtime/utils/schemaview.py
+++ b/linkml_runtime/utils/schemaview.py
@@ -1492,8 +1492,11 @@ class SchemaView(object):
             if k not in dest.types:
                 dest.types[k] = copy(v)
         for k, v in schema.enums.items():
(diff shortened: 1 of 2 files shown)


linkml_runtime/utils/schemaview.py
```diff
@@ -1492,8 +1492,11 @@ def merge_schema(self, schema: SchemaDefinition) -> None:
             if k not in dest.types:
                 dest.types[k] = copy(v)
         for k, v in schema.enums.items():
-            if k not in dest.types:
```
turbomam, line 1495: whoa, did you catch a bug?!

cmungall: yep, updated top comment!

PR #231, opened 2022-11-30
add the ability to declare "prefixmaps" contexts (including bioregistry contexts) in default_curie_maps.  fixes #604

sierra-moxon:
use case here is:
- project needs a new prefix to be added to Biolink
- create prefix in bioregistry 
   - not available in prefixcommons without secondary curation in prefixcommons
   - prefix isn't accessible via default_curie_maps in LinkML 
   - dual annotation required in Biolink

Incorporate prefixmaps context lookup functionality to linkml-runtime.utils.namespaces.add_prefixmap so that we can use bioregistry contexts as a fallback when none of obo_context, idot_contex, monarch_context, semweb_context, etc. contain the prefix. 

somewhat related to https://github.com/linkml/linkml/issues/852 - this doesn't deprecate default_prefix_maps, it just adds to the available maps that folks can use (incl. bioregistry maps).  

The next step here is to allow this changed method to read a prefix map directly without the requirement that it be in prefixcommons or prefixmaps.  

Hard-coding the map names here is not the best approach; I see a TODO in prefixmaps repo that those at least should be introspected from a config file (👍 ).

### DIFF
--- a/linkml_runtime/utils/namespaces.py
+++ b/linkml_runtime/utils/namespaces.py
@@ -241,10 +241,12 @@ class Namespaces(CaseInsensitiveDict):
 
         if map_name == 'prefixmaps':
             context = load_multi_context(["merged"])
-            prefix_map = context
+            prefix_map = context.as_extended_prefix_map()
+            print(type(prefix_map))
             print(prefix_map)
         else:
             prefix_map = curie_util.read_biocontext(map_name)
+            print(type(prefix_map))
             print(prefix_map)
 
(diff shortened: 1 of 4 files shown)


caufieldjh:
Taking a look - is the plan to use a list of prioritized prefix contexts as fallbacks or just one context?

sierra-moxon:
in biolink-model, I want to use `merged` alone from prefixmaps.

cmungall (approved):

PR #232, opened 2022-12-09
Update to latest metamodel

cmungall:


pkalita-lbl:
@cmungall I can't add you as a reviewer on this because you're also an author. I copied over the latest generated artifacts from `linklm-model`. Look good to you?

PR #235, opened 2022-12-21
Add methods to replace functionality in biolink model toolkit with schemaview

sierra-moxon:
This includes two convenience methods: 
1) all_aliases (a method to return all aliases from any schema element), used in a [method in BMT](https://github.com/biolink/biolink-model-toolkit/blob/4f93dd1be7327bcad3654283856c1ef5872413ae/bmt/toolkit.py#L443) to check which element a user is looking for when they ask by name.
2) get_children (for any element, not specifically a class, slot, etc.)
https://github.com/biolink/biolink-model-toolkit/blob/4f93dd1be7327bcad3654283856c1ef5872413ae/bmt/toolkit.py#L368

### DIFF
--- a/linkml_runtime/utils/schemaview.py
+++ b/linkml_runtime/utils/schemaview.py
@@ -602,9 +602,10 @@ class SchemaView(object):
         """
         children = []
         for e, el in self.all_elements().items():
-            if "is_a" in el and el["is_a"] == name:
-                children.append(el.name)
-            if "mixins" in el and mixin and name in el.mixins:
+            if isinstance(el, (ClassDefinition, SlotDefinition, EnumDefinition)):
+                if el.is_a and el.is_a == name:
+                    children.append(el.name)
+            if el.mixins and name in el.mixins:
                 children.append(el.name)
         return children
 
--- a/linkml_runtime/utils/schemaview.py
+++ b/linkml_runtime/utils/schemaview.py
@@ -946,7 +946,7 @@ class SchemaView(object):
         aliases = []
 
         for e, el in self.all_elements().items():
-            if "aliases" in el and el.aliases is not None:
+            if el.aliases and el.aliases is not None:
                 for alias in el.aliases:
                     aliases.append(alias)
         return aliases
--- a/linkml_runtime/utils/schemaview.py
+++ b/linkml_runtime/utils/schemaview.py
@@ -593,6 +593,22 @@ class SchemaView(object):
             return []
 
     @lru_cache()
+    def get_children(self, name: str, mixin: bool = True) -> List[str]:
+        """
+        get the children of an element (any class, slot, enum, type)
+        :param name: name of the parent element
+        :param mixin: include mixins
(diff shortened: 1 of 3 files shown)


linkml_runtime/utils/schemaview.py
```diff
@@ -592,6 +592,22 @@ def type_parents(self, type_name: TYPE_NAME, imports=True) -> List[TypeDefinitio
...
+    def get_children(self, name: str, mixin: bool = True) -> List[str]:
+        """
+        get the children of an element (any class, slot, enum, type)
+        :param name: name of the parent element
+        :param mixin: include mixins
+        :return: list of child element
+        """
+        children = []
+        for e, el in self.all_elements().items():
+            if "is_a" in el and el["is_a"] == name:
```
cmungall, line 605: I prefer not accessing the dict directly

e.g.

```
if isinstance(el, (ClassDefinition, SlotDefinition, EnumDefinition)):
  if el.is_a == name:
...
```

note also TypeDefinition is not totally symmetric, it uses typeof
```

File: linkml_runtime/utils/schemaview.py [hunk 2]
```diff
@@ -919,6 +935,21 @@ def get_elements_applicable_by_prefix(self, prefix: str) -> List[str]:
...
+    def all_aliases(self) -> List[str]:
+        """
+        Get the aliases
+
+        :return: list of aliases
+        """
+        aliases = []
+
+        for e, el in self.all_elements().items():
+            if "aliases" in el and el.aliases is not None:
```
cmungall, line 948: see above

linkml_runtime/utils/schemaview.py
sierra-moxon: fixed.

linkml_runtime/utils/schemaview.py
sierra-moxon: fixed

linkml_runtime/utils/schemaview.py
```diff
@@ -919,6 +935,21 @@ def get_elements_applicable_by_prefix(self, prefix: str) -> List[str]:
 
         return applicable_elements
 
+    @lru_cache()
+    def all_aliases(self) -> List[str]:
+        """
+        Get the aliases
```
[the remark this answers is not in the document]
sierra-moxon: added - I ran into something sort of magical.  I can't specify `alias_predicate` in my test data, instead because the slot `alias_predicate` has a `linkml:alias` slot, `predicate`, I can only use `predicate` in my test data. 
https://github.com/linkml/linkml-model/blob/61a6ad31ea5650b17e0f2d51b8497a3aff34e7cb/linkml_model/model/schema/meta.yaml#L401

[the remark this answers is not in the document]
sierra-moxon: perhaps we should remove the `alias: predicate` from the slot `alias_predicate` ?

cmungall (approved):

PR #236, opened 2022-12-23
Proposed update for https://github.com/linkml/linkml/issues/1188

hsolbrig:
Note that the test code in test_loaders_dumpers/test_loaders_dumpers.py/test_load_from_list changed.  Needs discussion


linkml_runtime/utils/yamlutils.py
```diff
@@ -93,6 +93,15 @@ def _is_empty(v: Any) -> bool:
         # TODO: Deprecate this function and migrate the python generator over to the stand alone is_empty
         return is_empty(v)
 
+    def _normalize_assignment(self, data: Any, target_class: Type) -> None:
```
cmungall, line 96: would it be possible to separate this change into another PR?

We are discussing the need for further normalization over here: https://github.com/linkml/linkml/pull/1189

All the other changes in this PR look ready to go, but I think the issue of normalization needs further discussion

sierra-moxon:
The conflicts in this PR with main are very difficult to resolve (I got most of them resolved, but the poetry.lock file is messy).  Should we recreate this change set on a more recent version of main?

cmungall:
Let’s just close this

PR #240, opened 2023-01-28
Check inlined or inlined_as_list slots only if range is class in schema

pkalita-lbl:
The root cause of https://github.com/linkml/linkml/issues/1266 is that the schema includes a slot defined like so:

```yaml
slots:
  see_also:
    range: uriorcurie
    description: Links to useful info
    multivalued: true
    inlined_as_list: true
```

When `JsonSchemaGenerator` switched from using `SchemaLoader` to `SchemaView` it introduced a call to `SchemaView.is_inlined()`. Currently that method will return `True` for the above slot definition. Regardless of whether you _should_ put `inlined_as_list` on a slot with a type range (as opposed to a class range), that feels like the wrong answer to me (I can already hear @cmungall saying "Postel-ian"). A `uriorcurie` can't really _be_ inlined because there's nothing _to_ inline, right?

The change here swaps the order of some logical checks in `is_inlined` so that the `inlined` and `inlined_as_list` metaslots are only checked if the range is a class in the schema. I added a test that attempts to enumerate all the possibilities. 

Fixes https://github.com/linkml/linkml/issues/1266

### DIFF
--- a/linkml_runtime/utils/schemaview.py
+++ b/linkml_runtime/utils/schemaview.py
@@ -1317,22 +1317,22 @@ class SchemaView(object):
         :param imports:
         :return:
         """
-        if slot.inlined:
-            return True
-        elif slot.inlined_as_list:
-            return True
-        else:
-            range = slot.range
-            if range in self.all_classes():
-                id_slot = self.get_identifier_slot(range, imports=imports)
-                if id_slot is None:
-                    # must be inlined as has no identifier
-                    return True
-                else:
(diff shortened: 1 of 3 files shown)


cmungall (approved):
Thanks, I agree with the analysis, and I approve the postelian invocation!

I think the spec should be tighter here:

- implementations MUST ignore inlined directives for types
- schemas SHOULD NOT declare type slots as multivalued

i.e this should generate warnings in future, but the behavior will always be tolerated

the most recent referencevalidator and specs go some way to addressing this

PR #245, opened 2023-02-06
1126 slot modifying classes

amc-corey-cox:
runtime portion to implement showing classes that modify a slot on the slot page for docgen.

https://github.com/linkml/linkml/issues/1126

### DIFF
--- a/tests/test_issues/test_linkml_issue_1126.py
+++ b/tests/test_issues/test_linkml_issue_1126.py
@@ -1,0 +1,28 @@
+import logging
+import unittest
+from unittest import TestCase
+from linkml_runtime.utils.schemaview import SchemaView
+from linkml_runtime.linkml_model import ClassDefinition
+
+from tests.test_issues.environment import env
+
+
+class Issue998TestCase(TestCase):
+    """
+    https://github.com/linkml/linkml/issues/1126
+    Test getting classes that modify a slot
+    """
+    env = env
+
+    def test_issue_1126_slot_classes(self):
+        view = SchemaView(env.input_path('linkml_issue_1126.yaml'))
+        slot_classes = view.get_classes_modifying_slot("type")
+        assert len(slot_classes) == 2
+        for element in slot_classes:
+            assert type(element) is ClassDefinition
+        assert slot_classes[0].name == "Programmer"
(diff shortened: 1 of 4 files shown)


amc-corey-cox:
Along with the other PR I think this should add the functionality we need.

amc-corey-cox:
As I mentioned in Slack. I had some issues with push this PR. As you can see with the last commit, I needed to fix an issue with the test_schemaview that was failing. We should verify this is an appropriate fix.

Separately, I was not able to rebase because doing so caused pytest to fail on gather tests from `tests/test_processing/test_referencevalidator.py`. These test failures appear to be due to failed imports in these lines.

`from examples import PermissibleValue`
`from linkml_runtime.utils.schema_builder import SchemaBuilder`

The first line can just be commented out but the second import issue is due to this file being missing from `main`:

`linkml_runtime/utils/schema_builder`

I wasn't sure how to reconcile that so I just created this PR without rebasing on `main`.

tests/test_utils/test_schemaview.py
```diff
@@ -702,7 +702,8 @@ def test_mergeimports(self):
                 "core", 
                 "prov", 
                 "xsd", 
-                "shex"], 
+                "shex",
+                "schema"],
```
cmungall, line 706: See https://github.com/linkml/linkml/issues/1285

cmungall: you took the right approach, but ideally this situation would not have happened

tests/test_issues/test_linkml_issue_1126.py
```diff
@@ -0,0 +1,28 @@
+import logging
+import unittest
+from unittest import TestCase
+from linkml_runtime.utils.schemaview import SchemaView
+from linkml_runtime.linkml_model import ClassDefinition
+
+from tests.test_issues.environment import env
+
+
+class Issue998TestCase(TestCase):
```
sujaypatil96, line 10: `class Issue998TestCase` or `class Issue1126TestCase`? Very trivial but just noticed it

sujaypatil96 (approved):
Logic that implements the retrieval of classes that modify certain slot definitions in `slot_usage` statements of that class looks excellent. Great work @amc-corey-cox.

PR #248, opened 2023-02-17
Imports to metamodel resolved locally rather than over network, #1102.

cmungall:
This solves the problem introduced in https://github.com/linkml/linkml/issues/502 but only for the metamodel.

See also https://github.com/linkml/linkml/issues/1102


cmungall:
@dalito any ideas on how to handle this? I think what is happening here is that when we obtain absolute paths on windows, it will make a path that looks like a CURIE, e.g. `D:\foo\bar`.

dalito:
I can have a look tomorrow. I think we had a very similar problem before.

PR #249, opened 2023-02-18
SchemaView initialization from a PosixPath

joeflack4:
Updates
```
- Update: UX: SchemaView: Creating a new instance now accepts a PosixPath rather than just a string representation of a file path.
```

### DIFF
--- a/linkml_runtime/utils/schemaview.py
+++ b/linkml_runtime/utils/schemaview.py
@@ -5,6 +5,7 @@ import collections
 from functools import lru_cache
 from copy import copy, deepcopy
 from collections import defaultdict, OrderedDict
+from pathlib import PosixPath
 from typing import Mapping, Tuple, Type, Union, Optional, List, Any
 
 from linkml_runtime.linkml_model import PermissibleValue, PermissibleValueText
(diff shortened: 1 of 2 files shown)


linkml_runtime/utils/schemaview.py
```diff
@@ -111,9 +112,9 @@ class SchemaView(object):
     modifications: int = 0
     uuid: str = None
 
-    def __init__(self, schema: Union[str, SchemaDefinition],
+    def __init__(self, schema: Union[str, PosixPath, SchemaDefinition],
                  importmap: Optional[Mapping[str, str]] = None, merge_imports: bool = False):
-        if isinstance(schema, str):
+        if isinstance(schema, str) or isinstance(schema, PosixPath):
```
joeflack4, line 117: I had 3 different proposed ways to do this in [this issue](https://github.com/linkml/linkml/issues/1056), but I felt like this way was even simpler than those.

joeflack4 (commented):
Addresses:
- https://github.com/linkml/linkml/issues/1056

sierra-moxon (approved):

sierra-moxon:
@joeflack4 - looks great!  if possible, could you add or edit a test to take advantage of the new support for PosixPath?

cmungall (changes requested):
Thanks!

I suggest generalizing this to work with other paths - see https://docs.python.org/3/library/pathlib.html - unless there is a reason to restrict?

joeflack4:
@cmungall I was going to look at these requests this weekend. Thanks for handling! I see as well the need to convert it to a string as well within `__init__`, thanks for handling that.

PR #251, opened 2023-03-01
[draft] CurieNamespace catalog that is not a singleton and that uses curies

kervel:
for feedback. don't feel confident here, looks way too complicated for what i'm trying to achieve.


kervel:
i just discovered we also have "namespacegen.py" which generates a python file that ... can be used to do expand/compress. so i have the impression that we have 100 ways to fix the "how to expand/compress uri's" problems.

so for an outsider it would be a bit strange to say look with namespacegen you can generate a Namespace object (which doesn't use the Namespaces class , which also can serve the same purpose btw), but if you then use pythongen we won't use that file to do exactly this.

capsulecorplab:
Should we perhaps open an issue for implementing a Namespace function that goes from uriorcurie to uri or curie?

amc-corey-cox:
@kervel  I have moved the work on this PR into the repo branch curienamespace_index_catalogclass . If you would like to continue this work please re-open the PR on the mono-repo after the merge from that branch. We'll lift it over for you so it should take less effort. This will have a slight rename in the linkml repo but you will be able to re-open it there.

PR #254, opened 2023-03-07
Fix get_classes_modifying_slot

amc-corey-cox:
Use slot name to check for classes modifying slot.

PR #255, opened 2023-03-14
linkml-model pre-release 1.5

cmungall:


cmungall:
for windows failure see discussion here #254

PR #256, opened 2023-03-22
pre-release of 1.5x series.

cmungall:
Replaces #255

Do not merge until we are ready to release 1.5


cmungall:
discussed on today's linkml call

PR #270, opened 2023-06-29
Regenerate python dataclasses for validation models

deepakunni3:
Currently the `linkml_runtime/linkml_model/validation.py` is broken because of missing imports.

This PR regenerates the python dataclasses for validation models defined in `linkml_runtime/linkml_model/model/schema/validation.yaml`.


sierra-moxon:
Thanks @deepakunni3 - is the auto regeneration of these part of the makefile and/or github actions?

deepakunni3:
I had to run them manually and then push to the branch. 

I did not see any github action that did autogeneration of artifacts in this repo.

Do you suggest I run gen-python for all the model YAML so that they are all in sync to the same run date?

PR #271, opened 2023-07-13
Ensure SchemaView uses importmap argument

cmungall:
fixes https://github.com/linkml/linkml/issues/1534

### DIFF
--- a/tests/test_utils/test_schemaview.py
+++ b/tests/test_utils/test_schemaview.py
@@ -353,6 +353,21 @@ class SchemaViewTestCase(unittest.TestCase):
         view.add_class(ClassDefinition('W'))
         self.assertCountEqual(['Y', 'Z', 'W'], view.all_classes())
 
+    def test_import_map(self):
+        """
+        Path to import file should be configurable
+        """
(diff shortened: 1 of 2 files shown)


tests/test_utils/test_schemaview.py
```diff
@@ -353,6 +353,21 @@ def test_caching(self):
...
         self.assertCountEqual(['Y', 'Z', 'W'], view.all_classes())
 
+    def test_import_map(self):
+        """
+        Path to import file should be configurable
+        """
+        for im in [{"core": "/no/such/file"}, {"linkml:": "/no/such/file"}]:
+            with self.assertRaises(FileNotFoundError):
+                view = SchemaView(SCHEMA_WITH_IMPORTS, importmap=im)
+                view.all_classes()
```
sujaypatil96, line 363: If the _importmap_ does not have existent paths to YAML files, then wouldn't it be better if the program errored out in the schema initialization step itself?

As in should we be waiting till the user makes a call to one of the SchemaView methods (in this case `all_classes()`) before they realize that they've provided an incorrect importmap?

```python
view = SchemaView(SCHEMA_WITH_IMPORTS, importmap=im)
```
itself should complain about non existent files.

PR #279, opened 2023-09-14
Issue 1627

pkalita-lbl:
Fixes https://github.com/linkml/linkml/issues/1627


pkalita-lbl:
@cmungall Can you double-check that these changes are still inline with what you were doing in https://github.com/linkml/linkml-runtime/pull/248

cmungall (approved):

PR #280, opened 2023-10-06
fix: avoid uri shadowing

Silvanoc:
URIorCURIE validation function is wrongly identifying as CURIEs certain URIs. This patch first tries to validate if a URIorCURIE is a valid URI, otherwise it tries to validate if it's a valid CURIE.


Silvanoc:
Following invalid **URIs** or **CURIEs** which weren't catched by the previous validation are not allowed anymore:

- Those having '[' or ']', like '[UCUM:mm[Hg]](https://github.com/linkml/linkml-runtime/pull/280/files#diff-506a228edb944f2ed4b8c39e9ed9d2e9001c44b03a4f8b74487e4692a77cff20L376)'. [^1]
- Those having other not allowed characters, like '[UO:0000000 ! unit](https://github.com/linkml/linkml-runtime/pull/280/files#diff-8c6e6f951d3a975272fe5e7d946cb408df2b4f3c34bd4917b9208eefd00673b8L113)'

Following invalid **URIs** which weren't catched by the previous validation are not allowed anymore:

- Those starting with [":"](https://github.com/linkml/linkml-runtime/pull/280/files#diff-7a490a7a4ef08a60bdbd15591839c8283b44be4600f74b2668eb0d0e937f7a50R86), though they remain being valid CURIEs.

Furthermore following valid **URIs** that were being rejected are allowed:

- Empty ones, like "[''](https://github.com/linkml/linkml-runtime/pull/280/files#diff-7a490a7a4ef08a60bdbd15591839c8283b44be4600f74b2668eb0d0e937f7a50R95)". It's a so-called same-document reference.
- Those that look like a CURIE, like "[rdf:type](https://github.com/linkml/linkml-runtime/pull/280/files#diff-7a490a7a4ef08a60bdbd15591839c8283b44be4600f74b2668eb0d0e937f7a50R97)". Though they remain being interpreted as CURIES within the scope of a `Uriorcurie` range.
- [URNs](https://github.com/linkml/linkml-runtime/pull/280/files#diff-7a490a7a4ef08a60bdbd15591839c8283b44be4600f74b2668eb0d0e937f7a50R98) are now also allowed. Like above, within the scope of a `Uriorcurie` range they remain being interpreted as CURIEs.

[^1]: Support for Safe-CURIEs, which might have '[' and ']' only surrounding the whole CURIE will be added on a later PR

Silvanoc:
Assuming the support for empty URIs (same-document reference) makes any difficulties, I don't think that deviating from the standard would harm much there.

Silvanoc:
@cthoyt the [module providing validation functions here](https://github.com/linkml/linkml-runtime/pull/280/files#diff-2a487e1dd6b4ed1ca17d0f143b82535326c3548a9b5a7b050a8c05d4dd0861c3) goes in the direction of what I was [wishing to see in the `curies` package](https://github.com/cthoyt/curies/issues/77). You might be interested on it or even review it.

Silvanoc:
Small hint for easy manual testing:
```
poetry run python -c 'from linkml_runtime.utils.uri_validator import validate_uri ; print(validate_uri("urn:namespace:resource"))'
```
will return a regex match object, since it succeeds.

Silvanoc:
I've added named groups for better unterstanding of how the validator is parsing the passed strings.

```
poetry run python -c 'from linkml_runtime.utils.uri_validator import validate_uri ; \
    from pprint import pprint ; \
    pprint(validate_uri("ldap://[2001:db8::7]/c=GB?objectClass?one").groupdict())'
```

returns

```
{'authority': '[2001:db8::7]',
 'fragment': None,
 'hier_part': '//[2001:db8::7]/c=GB',
 'host': '[2001:db8::7]',
 'port': None,
 'query': 'objectClass?one',
 'scheme': 'ldap',
 'uri': 'ldap://[2001:db8::7]/c=GB?objectClass?one',
 'userinfo': None}
```

Silvanoc:
@sierra-moxon are you taking care of assigning the PR or are you expecting me to pick someone?

cthoyt:
@Silvanoc I'm glad you found a place for this!

PR #281, opened 2023-10-31
Optimize implementation of SchemaView `get_classes_by_slot()` method

sujaypatil96:
The `get_classes_by_slot()` method in SchemaView takes an extremely long time to run on the MIxS schema and generate the _Applicable Classes_ table on slot documentation pages because of which we are having to explore ways to optimize the runtime for the `get_classes_by_slot()` method.

### DIFF
--- a/linkml_runtime/utils/schemaview.py
+++ b/linkml_runtime/utils/schemaview.py
@@ -1435,7 +1435,7 @@ class SchemaView(object):
             if x.range:
                 range_union_of.append(x.range)
         return range_union_of
-
+    
     def get_classes_by_slot(self, slot: SlotDefinition, include_induced: bool = False) -> List[ClassDefinitionName]:
         """Get all classes that use a given slot, either as a direct or induced slot.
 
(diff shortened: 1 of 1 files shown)


sierra-moxon:
We did some profiling (coincidentally :D) of docgen and found that the `deepcopy` calls for large schemas in schemaview were the most time-consuming bit.
https://github.com/linkml/linkml-runtime/blob/687fc53338cc791d0f932c3b9a22c0e8313fd99b/linkml_runtime/utils/schemaview.py#L1201
https://github.com/linkml/linkml-runtime/blob/687fc53338cc791d0f932c3b9a22c0e8313fd99b/linkml_runtime/utils/schemaview.py#L1209

sujaypatil96:
Oh, that's very good to know, thank you for digging into this @sierra-moxon 😁

cmungall:
good sleuthing! we should definitely avoid use of deepcopy when calculating induced slots

but note that induced slots should not be used for docgen purposes. We only want to show asserted usages. Otherwise we may end up flooding the user (e.g. in biolink, every named thing has `id` as an induced slot)

sierra-moxon:
changing https://github.com/linkml/linkml/blob/598376ce7f8c11bd3cf31f0ca7e3d5c34770021a/linkml/generators/docgen/slot.md.jinja2#L26 from True to False in my custom template (and other instances in this file) took docgen on biolink down from >2 hours to just under a minute.

sierra-moxon:
see also #1214 and #1604

cmungall:
@sujaypatil96 is this replaced by #300?

PR #285, opened 2023-11-08
Improve curie testing

Silvanoc:
Removing wrong expectations on CURIEs and adding new expectations. Also fixing validation according the specification.

### DIFF
--- a/tests/test_utils/test_metamodelcore.py
+++ b/tests/test_utils/test_metamodelcore.py
@@ -67,9 +67,8 @@ class MetamodelCoreTest(unittest.TestCase):
         with self.assertRaises(ValueError):
             Curie("1df:type")
         self.assertFalse(Curie.is_valid('1df:type'))
-        with self.assertRaises(ValueError):
-            Curie("rdf:17")
-        self.assertFalse(Curie.is_valid('rdf:17'))
+        self.assertTrue(Curie.is_valid('rdf:17'))
+        self.assertTrue(Curie.is_valid('17'))
         nsm = Namespaces(Graph())
         self.assertEqual(RDF.type, Curie("rdf:type").as_uri(nsm))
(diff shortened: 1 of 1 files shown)


cmungall:
You you still working on this PR @Silvanoc ?

Silvanoc:
> You you still working on this PR @Silvanoc ?

Sorry, I completely forgot this PR. Ich had a private discussion on Slack with @sierra-moxon, where she was raising her concern about the convenience of the changes implying that something like `17` would become an accepted CURIE (see the tests).

I'm quoting here her concern and my answer:

By @sierra-moxon:
> Hi again - I see you opened another draft PR on this -- and are currently validating that 17 is a valid curie?  I don't think that is a valid curie (I don't claim to be an expert here, but if I had data that used 17 as an identifier, I would want a validation exception generated).

By @Silvanoc:
> If I've read the spec right, 17 is a valid CURIE as abc. If LinkML should only support a subset of the spec-conform CURIEs, then it probably should be specified within the project. My feeling is that you expect at least the colon : and possibly also the prefix. We can of course have validation rules for the LinkML CURIE subset. But as I've said, we should first have a clear specification. We could have something like "LinkML CURIEs are all valid CURIEs according the W3C specification providing both a prefix and the separation colon".

After that I simply forgot about it.

@cmungall what's your opinion about it? Could you formulate it with comments on the tests? Something like "I wouldn't expect this to be accepted" or otherwise.

Silvanoc:
According [this issue on `curies`](https://github.com/biopragmatics/curies/issues/103) from @cmungall, you don't seem to be that happy with the W3C CURIEs standard. Still other people interested on LinkML (like the W3C [Web of Things](https://www.w3.org/WoT/) project) probably want to conform to the W3C standard. I wonder if we should make LinkML somehow capable of coping with both (using a flag like `--strict-curies` or similar).

PR #286, opened 2023-11-09
Add support for Python 3.12

Silvanoc:
Add test to validate that "greenlet" is on a version that makes the project runnable on Python v3.12.

Update the Poetry locking file to install such a version. Should someone try to downgrade "greenlet" to a version that cannot run on Python 3.12, the test will fail.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -12,7 +12,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest]
-        python-version: ["3.8", "3.9", "3.10"]
+        python-version: ["3.8", "3.9", "3.10", "3.12"]
         exclude:
         - os: windows-latest
           python-version: "3.8"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,9 +26,11 @@ classifiers = [
     "Intended Audience :: Healthcare Industry",
     "License :: CC0 1.0 Universal (CC0 1.0) Public Domain Dedication",
     "Topic :: Software Development :: Libraries :: Python Modules",
-    "Programming Language :: Python :: 3.7",
     "Programming Language :: Python :: 3.8",
-    "Programming Language :: Python :: 3.9"
+    "Programming Language :: Python :: 3.9",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: 3.12",
 ]
 
 packages = [
@@ -40,8 +42,7 @@ comparefiles = "linkml_runtime.utils.comparefiles:cli"
 linkml-normalize = "linkml_runtime.processing.referencevalidator:cli"
 
 [tool.poetry.dependencies]
-python = "^3.7.6"
-
+python = "^3.8"
 click = "*"
 deprecated = "*"
 hbreader = "*"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -59,7 +59,6 @@ pydantic = ">=1.10.2, <3.0.0"
 
 [tool.poetry.dev-dependencies]
 coverage = "^6.2"
-importlib-metadata = {version = "^4.12.0", python = "<3.8"}
 
 [build-system]
 requires = ["poetry-core>=1.0.0"]
--- a/linkml_runtime/__init__.py
+++ b/linkml_runtime/__init__.py
@@ -6,10 +6,7 @@ from rdflib import RDF, RDFS, SKOS, XSD, OWL
 # use importlib.metadata to read the version provided
 # by the package during installation. Do not hardcode
 # the version in the code
-try:
-    import importlib.metadata as importlib_metadata
-except ModuleNotFoundError:
-    import importlib_metadata
+import importlib.metadata as importlib_metadata
 
 LINKML = CurieNamespace('linkml', 'https://w3id.org/linkml/')
 TCCM = CurieNamespace('tccm', 'https://ontologies.r.us/tccm/')
--- a/linkml_runtime/utils/metamodelcore.py
+++ b/linkml_runtime/utils/metamodelcore.py
@@ -278,6 +278,8 @@ class XSDDate(str, TypedNode):
             value = value.value
         if isinstance(value, datetime.date):
             value = value.isoformat()
+        if not re.match(r'^\d{4}-\d{2}-\d{2}$', value):
+            return False
         try:
             datetime.date.fromisoformat(str(value))
         except ValueError:
--- a/poetry.lock
+++ b/poetry.lock
@@ -2,14 +2,14 @@
 
 [[package]]
 name = "annotated-types"
-version = "0.5.0"
+version = "0.6.0"
 description = "Reusable constraint types to use with typing.Annotated"
 category = "main"
 optional = false
-python-versions = ">=3.7"
+python-versions = ">=3.8"
 files = [
-    {file = "annotated_types-0.5.0-py3-none-any.whl", hash = "sha256:58da39888f92c276ad970249761ebea80ba544b77acddaa1a4d6cf78287d45fd"},
-    {file = "annotated_types-0.5.0.tar.gz", hash = "sha256:47cdc3490d9ac1506ce92c7aaa76c579dc3509ff11e098fc867e5130ab7be802"},
+    {file = "annotated_types-0.6.0-py3-none-any.whl", hash = "sha256:0641064de18ba7a25dee8f96403ebc39113d0cb953a01429249d5c7564666a43"},
+    {file = "annotated_types-0.6.0.tar.gz", hash = "sha256:563339e807e53ffd9c267e99fc6d9ea23eb8443c08f112651963e24e22f84a5d"},
 ]
 
 [package.dependencies]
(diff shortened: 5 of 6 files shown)


Silvanoc:
Hopefully the [same approach used for linkml](https://github.com/linkml/linkml/pull/1704#issuecomment-1809793719) fixes it here too.

pyproject.toml
```diff
@@ -40,8 +41,7 @@ comparefiles = "linkml_runtime.utils.comparefiles:cli"
 linkml-normalize = "linkml_runtime.processing.referencevalidator:cli"
 
 [tool.poetry.dependencies]
-python = "^3.7.6"
-
+python = "^3.9.0"
```
pkalita-lbl, line 44: Is there a reason this needs to be 3.9 instead of 3.8? Just curious since 3.8 won't reach end-of-life until [later this year](https://devguide.python.org/versions/), and I think we should try and keep this package as "installable" as possible.

Silvanoc: Upgrading Prefixmaps to a version that [requires Python to be 3.9 or higher](https://github.com/linkml/prefixmaps/blob/main/pyproject.toml#L27), implicitly implies that 3.8 won't be supported anymore.

That version of Prefixmaps is needed to have support for Python 3.12. Therefore as I see it, it's XOR decision between supporting 3.8 and supporting 3.12. I haven't found any other way to have support for Python 3.12.

Silvanoc:
> PR [linkml/prefixmaps#66](https://github.com/linkml/prefixmaps/pull/66) will fix the conflict between having support for both Python 3.8 and 3.12 simultaneously due to dependencies in that package. Therefore marking this PR as draft until a version of prefixmaps with the fix has been released, since we will then need to readapt this patch.

@pkalita-lbl as expected, your fix in prefixmaps has enabled adding support for Python 3.12 without sacrifycing support for Python 3.8. Tests cover both versions and succeed.

Silvanoc:
The issue being fixed by commit 90650c00f1b5e5d55810acebfadbac7fd927e2c7 (and being tested thanks to commit 65e043770f38126cbb053973ac7ede963c417f4e) has been detected and described in https://github.com/linkml/linkml/pull/1704.

It appeared in fact with Python 3.11, but tests were not being run against that Python version, the issue remained undiscovered. With the activation of tests agains Python 3.12 the issue has been uncovered.

pyproject.toml
```diff
@@ -52,13 +53,13 @@ prefixcommons = ">=0.1.12"
...
 requests = "*"
-prefixmaps = ">=0.1.4"
+prefixmaps = ">=0.2.2"
 curies = ">=0.5.4"
 pydantic = ">=1.10.2, <3.0.0"
 
 [tool.poetry.dev-dependencies]
 coverage = "^6.2"
-importlib-metadata = {version = "^4.12.0", python = "<3.8"}
+importlib-metadata = {version = "^4.12.0", python = "^3.8"}
```
sneakers-the-rat, line 62: i think this can get removed? or else if we depend on features specifically from `^4.12.0` it looks like we can do `python = "^3.10"`

sneakers-the-rat: (also lgtm but idk if i'm allowed to give you a thumbs up review lol)

sneakers-the-rat: what i would say is that for the `metadata()` call to get the version here, which is the one place that it's used: https://github.com/linkml/linkml/blob/7ea1a33226bb6480c7e0bea5b1c5ad11347dcceb/linkml/_version.py#L5

if we are going `>=3.8` then we can remove it, for its purpose has been served <3

it is useful in the sense that it forced me to implement a whole bunch of specific behavior for conditional imports i wouldn't have remembered to otherwise (but will inevitably be needed at some point) in this: https://github.com/linkml/linkml/pull/1927/files

so it gave us one last gift (and hints that we should probably take that out of the python generators too

Silvanoc: `importlib-metadata` can then be completely from `[tool.poetry.dev-dependencies]`, right?

sneakers-the-rat: ya and nab the place it's imported in `_version` and should be good?

pyproject.toml
Silvanoc: @pkalita-lbl can we mark this conversation as resolved? IMO the posterior changes on the branch to take advantage of your fixes in `prefixmaps` should be sufficient.

cmungall (approved):

Silvanoc:
AFAIK I'm no maintainer in this project, but the "Merge pull request" button is active.

Now I wonder if I'm expected to merge after having gotten the approval from a maintainer or rather wait for another maintainer to merge it 🤔

I would assume that the approval of a maintainer overrides the open conversations (equivalent to resolving all conversations). Is it so?

pyproject.toml
```diff
@@ -52,13 +53,12 @@ prefixcommons = ">=0.1.12"
 pyyaml = "*"
 rdflib = ">=6.0.0"
 requests = "*"
-prefixmaps = ">=0.1.4"
+prefixmaps = ">=0.2.2"
```
pkalita-lbl, line 56: We have the version of `prefixmaps` specified at `0.2.2` in `poetry.lock`. That's good! Our local dev environments and CI environments will be able to test under Python 3.8 through Python 3.12. But I think we _shouldn't_ change this specification in `pyproject.toml`. This change would mean all clients of `linkml` need to use a very recent version of `prefixmaps`, causing potential dependency incompatibilities. A client of `linkml` should be able to continue using `prefixmaps < 0.2.2` in their environment.

pkalita-lbl (commented):
Sorry for the delay in looking at this. I'd recommend not modifying the `prefixmaps` specification in `pyproject.toml` because we don't need to force _our_ testing environment needs on downstream clients.

pyproject.toml
sneakers-the-rat: Wasnt that update to fix some incompatibility between 3.8 and 3.12? So the dependency should be like >=0.1.4 for python<3.12 and >=0.2.2 for 3.12?

Silvanoc:
I don't fully know the "etiquette" in this project WRT open conversations. In other projects is the opener of a conversation responsible for clicking "Resolve conversation" once satisfied with the replies. How should I handle them here? Close them myself, when I consider I've already clarified all open points?

pyproject.toml
sneakers-the-rat: For some reason I am not seeing "resolve conversation" on my end, so if you do, then resolve it :)

.github/workflows/main.yaml
```diff
@@ -12,7 +12,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest]
-        python-version: ["3.8", "3.9", "3.10"]
+        python-version: ["3.8", "3.9", "3.10", "3.12"]
```
sneakers-the-rat, line 15: One last thing, why no 3.11 here?

cmungall:
We don't have this SOP formally defined. Is there a good template we can
follow? If so we can document here
https://linkml.io/linkml/contributing/contributing.html

For now, use your best judgment (and we should figure how to make it such
that you can resolve comments you yourself made)

.github/workflows/main.yaml
Silvanoc: Just because it wasn't there before and I didn't want to create additional testing load, when my focus was 3.12 and not 3.11 (which was already supported).

Though on the other side [I've already found an issue raising on `linkml`](https://github.com/linkml/linkml/pull/1704#issuecomment-1952810752) because `linkml-runtime` wasn't being tested on 3.11. I will make sure to add the test for 3.11 on this [follow-up PR](https://github.com/linkml/linkml-runtime/pull/299). I will add it also once we have an agreement on how to resolve [this conversation](https://github.com/linkml/linkml-runtime/pull/286#pullrequestreview-1894203188).

cmungall: I'm all for incremental, so I'll resolve this conversation, but feel free to make a new PR when this is done (I don't like extra testing load but linkml-runtime is quite light compared to linkml)

sneakers-the-rat: Ig if we are testing on either side of it, should be fine???

pyproject.toml
Silvanoc: To avoid the long feedback round-trips due to time-shift, I'm applying my proposal on the branch with the hope that it's addressing the concerns of this conversation. I'm not merging though before getting the confirmation from @pkalita-lbl

pkalita-lbl: Sorry I'm not trying to drag this out, but this is still over-constraining. If a client is using Python 3.12 there's no reason they need to also have `prefixmaps >= 0.2.2`. Versions of `prefixmaps` going back to 0.2.0 declare support for Python 3.12. All `linkml` needs to do is say "we need `prefixmaps >= 0.1.4`" as we did before. If a client is using Python 3.12, it's up to their package installation tool (whether that's pip or Poetry or whatever) to resolve an appropriate version of `prefixmaps` for their environment.

When it comes to _testing_, we will get a version (0.2.2) that's compatible with all the versions of Python we want to test with because of what's specified in `poetry.lock`.

Silvanoc: Now I got your point.

Even if `linkml-runtime` declares that `prefixmaps` should be "only" `>=0.1.4` on Python 3.12, the `prefixmaps` package itself already does so. It means that, the moment that Pip or Poetry tries to resolve dependencies, it realizes that it needs `prefixmaps` `>=0.2.2`. You're fully right!

I'm dropping my last commit and removing the change in the `pyproject.toml` WRT `prefixmaps`.

pkalita-lbl: Yes you got it! I'll keep an eye out for that commit so we can finally merge this work!

Silvanoc: It should be done now. Please have a look at it and feel free to either merge yourself or approve (then I'd merge).

@pkalita-lbl thanks for your valuable input!

sneakers-the-rat: i'm also not trying to be be difficult, and ultimately think >=0.1.4 would be sufficient in any first-order dependency situation by virtue of how `pip` and `poetry` will both resolve to the most recent version available and not try and install `<0.2`, but since `prefixmaps<2` declare python compatibility `^3.7.6` it is still possible for a 3.12 project to resolve to `prefixmaps<2`. This would be a problem with a package that depends on `linkml` and any other project that specifies `<0.2` (or `^0.1.*`, etc.). Thinking about bioregistry and pydantic 1/2 problems here. 

arguably the ideal thing would be for `prefixmaps<0.2` to specify `python<3.12` if it's known to not work, but unless we want to issue a bunch of `postN` releases (which i hope we don't) it doesn't. but if we know that `<0.2` doesn't work for `3.12` then we should spec that here - dependency resolution should correctly fail for a project that tries to use `prefixmaps<0.2` in 3.12. 

The role of version specifiers is to exclude combinations of dependencies known to be incompatible through an entire dependency tree - we could be maximally unrestrictive and let people YOLO their own environments entirely and leave all version specifiers blank, but then suffer the consequences of everyone's environment only working by chance. IMO it's good practice to exclude known bad dep combos, in this case we're not excluding any valid environments, but preventing unforseen footguns.

ultimately not my call! just looking out for future issue triagers <3

cmungall (approved):

pyproject.toml
sneakers-the-rat: oh sorry i didn't refresh my page and see that you had responded and also resolved this thread. my bad! whatevs, i'm still lgtm and maybe this is a more general discussion about how to manage deps

pkalita-lbl: I totally understand what you're saying.

Just stepping back, in the past I think `linkml` has erred (intentionally or not) on the side of being too restrictive in its dependency specifications. And it has led to downstream clients reporting the inability to install `linkml` into their environment even though ostensibly it shouldn't be a problem (if, that is, our dependency spec wasn't so strict). Because of that history I'm just trying be very intentional about this kind of change.

For this specific issue I'd just point out that:

> if we know that <0.2 doesn't work for 3.12 then we should spec that here

We don't know that, not at all. When I said that versions of `prefixmaps` declare support for Python 3.12 starting in `0.2.0` that was just based on a cursory glance of the `classifiers` field in their `pyproject.toml` in recent version tags. And my point was just that requiring `prefixmaps` >=0.2.**2** on Python 3.12 was _definitely_ too restrictive.

> IMO it's good practice to exclude known bad dep combos

That's totally fair! But I think there is no _known_ bad combo here. I might have miscommunicated and caused some confusion there. If so, my bad.

sneakers-the-rat: oh gotcha, i assumed we knew there was some problem with 3.12 just skimming the previous discussion on this PR. ya if we don't know there's an incompatibility, seems good to me!

cmungall:
Looks like this is also bundling in a minimum requirement of pydantic to 2.5 - not against this per se but I think this should be made into a separate PR?

PR #288, opened 2023-12-01
check that install of "latest dependent packages" of linkml_runtime still results in tests passing (as opposed to running tests only on the poetry.lock dependencies)

sierra-moxon:
add a check dependencies action that removes the poetry.lock file and reinstalls the linkml-runtime environment (getting the latest compatible packages in the pyproject.toml specification) and then runs the tests as a way of testing the code in a PR on the latest package dependencies

partially resolves #1749.  

After discussion during developer days hackathon, we want to limit our downstream build dependencies (e.g. not build out bmt, etc from LinkML packages directly), but makes sense to take this first step towards "latest possible" dependency checking.

### DIFF
--- a/.github/workflows/check-dependencies.yaml
+++ b/.github/workflows/check-dependencies.yaml
@@ -1,8 +1,4 @@
-# Built from:
-# https://docs.github.com/en/actions/guides/building-and-testing-python
-# https://github.com/actions/setup-python/
-
-name: Build and test linkml-runtime
+name: Build and test linkml-runtime package dependencies
 
 on: [pull_request]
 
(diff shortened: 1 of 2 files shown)


pkalita-lbl (commented):
The more I think about this the more hesitant I am about it. 

Stepping back from https://github.com/linkml/linkml/issues/1749 specifically, there is a question that comes up from time to time about whether we should be committing our `poetry.lock` file (related: https://github.com/linkml/linkml/issues/1641, https://github.com/linkml/linkml/issues/1567). The [Poetry documentation](https://python-poetry.org/docs/basic-usage/#committing-your-poetrylock-file-to-version-control) lays it out the concerns pretty concisely. `linkml-runtime` is clearly a library, and so there's no cut-and-dried answer. We have to consider the trade-offs between reproducibility (by including the lock file) and faster failures due to new dependency versions (by omitting the lock file). Given the limited resources we have to devote to LinkML, we've opted for reproducibility.

I'm concerned that these changes effectively reverse that decision. I'm having hard time imaging a scenario in which this new workflow (running tests after installing _without_ the lock file) passes but the old one (running tests after installing _with_ the lock file) fails. So in that sense why even keep the old one? And in that sense why even commit the lock file? It's totally valid for us to reconsider our stance on the lock file, but we should probably do so in a wider discussion than this PR.

Coming back to https://github.com/linkml/linkml/issues/1749 it isn't even clear to me how this new test workflow could have prevented that situation. While the issue is written to suggest we're lacking testing infrastructure, I don't think that's true. We have the infrastructure; we simply failed to write the proper test case (ahead of time anyway; it has been addressed now) that would have failed and therefore prevent the offending change from being released. Again, I'm happy to reconsider the lock file issue, but we should frame the discussion correctly and I don't think https://github.com/linkml/linkml/issues/1749 is the right framing

sierra-moxon:
I agree it's a little bit pushy w/re to upgrade if this action fails.  I did use the methodology downstream in `biolink-model` to find a breaking change in `linkml 1.6.4` that wasn't caught by the existing linkml tests, nor was it caught by the poetry.lock version of the biolink-model repo (as that was still at `linkml 1.6.3`).

PR #292, opened 2023-12-18
Issue 1749 gh actions - add once a week dependency check

sierra-moxon:


.github/workflows/check-dependencies.yaml
```diff
@@ -0,0 +1,67 @@
+name: Build and test linkml-runtime package dependencies
```
pkalita-lbl, line 1: I'm not sure this accurately expresses what this is doing. Maybe change this to something like "Build and test linkml-runtime with latest dependency versions"?

.github/workflows/check-dependencies.yaml
```diff
@@ -0,0 +1,67 @@
...
+  # Allows you to run this workflow manually from the Actions tab
+    types: trigger-run-check-dependencies
+
+jobs:
+  test:
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, windows-latest]
+        python-version: ["3.7", "3.8", "3.9", "3.10"]
```
pkalita-lbl, line 16: Off-topic but there's not need to test 3.7 anymore. It has been EOL [for a little while now](https://devguide.python.org/versions/).

sierra-moxon: removed from two github actions

.github/workflows/check-dependencies.yaml
sierra-moxon: sure.

pkalita-lbl (approved):

PR #294, opened 2024-01-08
remove slot name redundancy in get_slots_by_enum method

sierra-moxon:
to avoid doc displays like this: 
https://biolink.github.io/biolink-model/ChemicalOrGeneOrGeneProductFormOrVariantEnum/
(`Slots` section)

While I was in there, I also converted the tests associated with this method to pytest and added the schema as a fixture.


sujaypatil96 (approved):
LGTM 🚀 removes redundant slots from being returned by `get_slots_by_enum()` method.

PR #300, opened 2024-02-20
optimize `get_classes_by_slot()` in `schemaview.py`

sujaypatil96:
The motivation to optimize this PR is that we are having to remove the _Applicable classes_ section in the generated slot markdown/web documentation pages: https://github.com/linkml/linkml/blob/main/linkml/generators/docgen/slot.md.jinja2#L36-L46 for large schemas like mixs.

The optimizations that have been applied here are:
* Use `set()` for accumulating class names
* reduce repetitive calls to `self.all_classes()`
* reduce number of iterations over _all_classes_
* direct/induced handled in separate loops

### DIFF
--- a/linkml_runtime/utils/schemaview.py
+++ b/linkml_runtime/utils/schemaview.py
@@ -1487,29 +1487,32 @@ class SchemaView(object):
             if x.range:
                 range_union_of.append(x.range)
         return range_union_of
-
-    def get_classes_by_slot(self, slot: SlotDefinition, include_induced: bool = False) -> List[ClassDefinitionName]:
+        
+    def get_classes_by_slot(
+        self, slot: SlotDefinition, include_induced: bool = False
+    ) -> List[ClassDefinitionName]:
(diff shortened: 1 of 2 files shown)


sujaypatil96:
After wonderful optimizations by Jonny in https://github.com/linkml/linkml-runtime/pull/296 I thought this would be an appropriate time to take advantage of his modifications and apply it to larged schemas (like mixs) that need this performance improvement.

sujaypatil96:
Still a work in progress, will open it up for review when ready.

cmungall (approved):

sujaypatil96:
@cmungall thanks for approving this PR. I'm still benchmarking if it's easier/faster to return `ClassDefinition` instead of `ClassDefinitionName` (corresponding changes will be required in slot.md.jinja2).

I won't merge this in till I have the results from the benchmarking.

sujaypatil96:
I think we are good to go with this PR as is @cmungall.

PR #306, opened 2024-03-13
[perf] Use yamllib if available

sneakers-the-rat:
Loading yaml is actually really slow in pure python! and we load a lot of yaml!

There's a free perf boost we can get by just using the compiled yamllib versions if they are available (they usually are). 

The only failure I noticed was that there was a single `\t` character at the start of a multiline string in the biolink model that cause s parsing error, but once that was fixed it was exactly the same just way faster.

For `YAMLLoader.load_as_dict` in main `linkml` tests, (cumulative time in seconds, `--with-slow`)

|                            | Total test time | `load_as_dict` | per call |
| ---------------- | -------------- | --------------- | ------- |
| Before PR | 1527s  | 451.3s | .1136s |
| This PR              | 1334s   | 243.9s | .062s |
| Change. | -193s | -207s | -.0516 |
Change:  | -12.6% | -46% | |

all ~ for free ~

### DIFF
--- a/linkml_runtime/utils/yamlutils.py
+++ b/linkml_runtime/utils/yamlutils.py
@@ -14,6 +14,11 @@ from linkml_runtime.utils.formatutils import is_empty
 
 YAMLObjTypes = Union[JsonObjTypes, "YAMLRoot"]
 
+try:
+    from yaml import CSafeLoader as SafeLoader
+except ImportError:
+    from yaml.loader import SafeLoader
+
 
 class YAMLMark(yaml.error.Mark):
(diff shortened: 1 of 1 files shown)


cmungall:
thx!

Looks like one failing test

We also want to consider whether to incorporate ruamel.yaml into our strategy. We use it in some places as it has nice features like comment preservations (very useful for diff/patch workflows). Not sure how it compares on performance

 https://yaml.readthedocs.io/en/latest/pyyaml/

cmungall (approved):

PR #307, opened 2024-03-13
[perf] Use local schemas if available

sneakers-the-rat:
Edit to add relation to prior issues:

(Partial)
Fix: https://github.com/linkml/linkml/issues/866
(Full)
Fix: https://github.com/linkml/linkml/issues/1012

And i also see this was handled in a few different ways in different prior PRs/issues:
- https://github.com/linkml/linkml-runtime/pull/248/
- https://github.com/linkml/linkml/issues/936
- https://github.com/linkml/linkml/issues/939
- https://github.com/linkml/linkml/pull/940
- etc.

This, to me, points to a greater need to simplify and unify the loading behavior, since it seems like we have a patchwork of fixes here that didn't quite reach the root of the problem because the loading behavior is quite complex. 

One can validate that network requests are still being made by, well, monitoring network traffic, as well as adding a debug flag just before the `hbread` printing what it's about to read.

---

Finally took the time to see what network requests were still happening during normal usage, because i kept hanging both on test runs and also when just trying to use the tool. 

Turns out that `hbread` doesn't use `requests` (which would be cached during testing) and just directly calls urllib. It also turns out that most of the time we are just requesting `types.yaml` over and over again, and so we can safely use the local version of the meta schema instead - our local version should always be the one we prefer, since it's tagged to the particular version of `linkml_runtime` that we're using, as opposed to the URI version which could be any version (ie. would be the most recent version even if we wanted to use an older version of the spec).

edit: this was removed to satisfy a test that needed the fileinfo:
> The only thing that looks dicey in here is me also trying to just cache all identical results from reads for a given cache period, and so I am avoiding passing `FileInfo` to `hbread` so that we can use `lru_cache` which requires hashable args. 

Perf of `request.py(urlopen)`:
Before: 288.5s (cumulative) 0.8291s per call
This PR: 18.94s (cumulative) 0.789s per call (we make fewer calls is the point)
Difference: -269s (-93%)

Edit: i have no idea why this test is failing - I tried to fix the source schema file and remove the newline at the end of the file, but otherwise i have no idea why it decided to stop printing the filename between 3 hours ago and now. i'll come back in the morning


cmungall:
Thanks! Yes, we should be phasing out the older hbread code over time

cmungall (approved):

PR #308, opened 2024-03-15
Fix namespaces CURIE prefix & suffix extraction

vincentkelleher:
## Motivation Behind This

While working on the LinkML JSON-LD Context generator, I chose to use the [namespaces.py `prefix_suffix(..)` method](https://github.com/linkml/linkml-runtime/blob/main/linkml_runtime/utils/namespaces.py#L192) to build `@id` attributes.

Then, I started having issues with the [test_curie_prefix_matching.py](https://github.com/linkml/linkml/blob/main/tests/test_issues/test_curie_prefix_matching.py) test class (BTW thanks to the developer of this class 👍 ) which has very specific CURIEs like `p1:/c/c1` for example.

## Solution

I went through the [CURIE specification](https://www.w3.org/TR/curie/) to find which types of CURIEs should be supported and I added them to the `test_namespaces.py` test class to make sure specific cases were managed.

The solution I propose is straightforward but as this method is used in many places, the impacts can be big. I have a pull request coming up to update the LinkML project with my JSON-LD Context generator developments and test updates.

### DIFF
--- a/linkml_runtime/utils/namespaces.py
+++ b/linkml_runtime/utils/namespaces.py
@@ -137,7 +137,9 @@ class Namespaces(CaseInsensitiveDict):
     def curie_for(self, uri: Any, default_ok: bool = True, pythonform: bool = False) -> Optional[str]:
         """
         Return the most appropriate CURIE for URI.  The first longest matching prefix used, if any.  If no CURIE is
-        present, None is returned
+        present, None is returned.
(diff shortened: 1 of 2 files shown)


cmungall (approved):

vincentkelleher:
Thanks @cmungall 🙏

PR #309, opened 2024-03-15
Update Cache To Support > 128 Entries

timothy-trinidad-ps:
Fixes https://github.com/linkml/linkml/issues/1973.

I'm starting with a really naive fix of just removing the entry limit for the `lru_cache`, but I'm open to pivoting to other solutions. 

With this change, running `gen-doc` on the following yaml file goes down from 3+ minutes to 7s.

[test.yaml.txt](https://github.com/linkml/linkml-runtime/files/14619348/test.yaml.txt)

### DIFF
--- a/linkml_runtime/utils/schemaview.py
+++ b/linkml_runtime/utils/schemaview.py
@@ -155,7 +155,7 @@ class SchemaView(object):
     def __hash__(self):
         return hash(self.__key())
 
-    @lru_cache()
+    @lru_cache(None)
     def namespaces(self) -> Namespaces:
         namespaces = Namespaces()
         for s in self.schema_map.values():
@@ -207,7 +207,7 @@ class SchemaView(object):
         schema = load_schema_wrap(sname + '.yaml', base_dir=base_dir)
         return schema
 
-    @lru_cache()
+    @lru_cache(None)
     def imports_closure(self, imports: bool = True, traverse: Optional[bool] = None, inject_metadata=True) -> List[SchemaDefinitionName]:
         """
         Return all imports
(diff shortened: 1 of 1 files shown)


sierra-moxon:
@timothy-trinidad-ps - looking lru doc, it seems like the default is "128" items in the cache and this PR takes it to "limitless" items in the cache.  

Do you think it would be helpful to take this item-number up in increments of 100 or so until we see a significant speedup without going limitless?

timothy-trinidad-ps:
I'm open to it, but any other number seemed just as arbitrary to me as 128. If I up it to 256 I know that we will run into this problem again in the next 6 months (we're in the process of annotating our database tables).

The obvious downside to this would be memory leaks for very large schemas, but I'm not entirely sure how to measure that. I think memory usage growing proportional to the schema size would be more expected than an arbitrary threshold after which the program effectively stops working.

sierra-moxon:
I can agree with that :) - is there anything we can do to prevent a very large schema from running into problems with too big a cache?  Said another way, what is the largest schema size you've tested this on so far?

timothy-trinidad-ps:
Just this 129. I can automate creating a schema with 1024 tables, but the specific use case I'm testing with is with `gen-doc` and the functions called by its dependent templates. What I would need help with is determining how to test other use cases of `schemaview.py` when it comes to memory usage.

timothy-trinidad-ps:
I just ran a benchmark for a file with 2,048 classes and 10,240 slots - it took about 4 minutes a peak of about 450 MB memory (at least for the `gen-doc` use case).
[test.yaml.txt](https://github.com/linkml/linkml-runtime/files/14620308/test.yaml.txt)

Python_47340

cmungall (approved):

PR #310, opened 2024-03-18
Fix `linkml_files`

sneakers-the-rat:
on windows, the paths for local model files look [like this](https://github.com/linkml/linkml/actions/runs/8305222724/job/22731836059?pr=1976#step:11:330):

```
'D:\\a\\linkml\\linkml\\.venv\\lib\\site-packages\\linkml_runtime\\linkml_model\\model/schema/types.yaml'
```

which seems to work, but it probably shouldn't. 

change hardcoded `/` separators to use `os.path.join` to make paths correct on windows

Related to: https://github.com/linkml/linkml/pull/1976

### DIFF
--- a/tests/test_linkml_model/test_linkml_files.py
+++ b/tests/test_linkml_model/test_linkml_files.py
@@ -1,0 +1,125 @@
+import pdb
+
+import pytest
+import requests
+from pathlib import Path
+from itertools import product
+from urllib.parse import urlparse
+
+from linkml_runtime.linkml_model.linkml_files import (
+    Source,
+    Format,
+    _Path,
+    URL_FOR,
+    LOCAL_PATH_FOR,
+    LOCAL_BASE,
+    GITHUB_IO_PATH_FOR,
+    GITHUB_PATH_FOR,
+    META_ONLY,
+    ReleaseTag
+)
+
+EXPECTED_FORMATS = [
+    (source, fmt) for source, fmt in product(Source, Format)
+     if (fmt not in META_ONLY and source != Source.META)
+]
+
+@pytest.mark.parametrize(
+    'source,fmt',
+    EXPECTED_FORMATS
+)
+def test_local_paths(source, fmt):
+    a_path = Path(LOCAL_PATH_FOR(source, fmt))
+    assert a_path.exists()
+    assert a_path.is_absolute()
+
+@pytest.mark.parametrize(
+    'fmt',
+    Format.__iter__()
+)
+def test_format_paths(fmt):
+    """Every format should have an entry in _Path"""
+    assert fmt.name in _Path.items()
+
+def test_no_unmapped_dirs():
+    """
+    There should be no additional directories that don't have a mapping for Format.
+    """
+    EXCLUDES = ('__pycache__',)
+
+    expected = {LOCAL_BASE / _Path.get(fmt.name).path for fmt in Format}
+    expected.add(LOCAL_BASE / 'model')
+
+    actual = {a_dir for a_dir in LOCAL_BASE.iterdir() if a_dir.is_dir() and a_dir.name not in EXCLUDES}
+    # Special case the root directory
+    actual.add(LOCAL_BASE)
+    # Special case YAML which is in a subdirectory - we've checked for existence above
+    actual.add(LOCAL_BASE / _Path.get('YAML').path)
+    assert expected == actual
+
+
+# --------------------------------------------------
+# URLs
+# --------------------------------------------------
+
+@pytest.mark.skip('We need to cache this...')
+@pytest.mark.parametrize(
+    'release_type',
+    ReleaseTag.__iter__()
+)
+@pytest.mark.parametrize(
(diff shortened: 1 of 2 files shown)


cmungall (approved):
thx!

I think we should use pathib over `os.path` but either is infinitely better than hardcoding slash

sneakers-the-rat:
> I think we should use pathib over os.path but either is infinitely better than hardcoding slash

just trying to follow the surrounding conventions :) ez to change

sneakers-the-rat:
OK i went to swap out `os.path.join` with `pathlib`, and when i went to write a simple test that the generated paths exist, it turns out that most of them don't and there were a lot of other problems with the module, and since this is something i'd like to be able to rely on i went ahead and rewrote most of it.

Problems:
- Many of the paths were wrong, the directory structure changed, particularly for the github paths.
- Several formats were missing - in the generated output, but not in the format mapping
- HTML didn't exist
- you can't have an enum that has multiple items with the same value, they end up shadowing each other, so the `Format` and `Path` enum didn't work.
- several of the formats only have `meta`, but there was no indication of that
- the URL building methods were mixed up with the path building methods

So this PR
- swaps out os.path.join with pathlib
- Replaces `Format` with an `Enum` that keeps all the items
- Moves the extension specification out of `Format` and into `_Path`
- Specifies directories and extensions with `FormatPath` within `_Path`
- Return PathParts from `_build_path` rather than a string, because paths and URLs need to be built differently
- corrects github paths 
- corrects github_io paths
- indicates which formats only generate META
- tests all of the above.

You'll notice a lot of skipped tests. those are the ones that make network requests, which we need to cache in the same way we do with `linkml`. I noticed this repo is still using `unittest` in its CI, and didn't want to swap that out here too, so whenever we switch to pytest and add `requests_cache` those tests shoudl automatically switch back on.

sneakers-the-rat:
ope i guess the tests aren't being run? idk how unittest works. ¯\_(ツ)_/¯ if ya want 'em they're there!

tests/test_linkml_model/test_linkml_files.py
```diff
@@ -0,0 +1,131 @@
+import pdb
```
cmungall, line 1: did you mean to leave this in?

cmungall:
Thanks!

I'd like to test this with the main `linkml` project before merging into main, as there are possibly some assumptions on the enum values in `linkml.__init()__` and possibly other places.

I realize the current SOP for these kinds of tests is suboptimal, so we can strategize ways to improve things.

I also realize I'm being very conservative here because as you point out a lot of this is legacy and things didn't break when paths changed..

I'm pretty certain this PR represents a much better way of doing things than before. But another approach might be to eliminate dependency on this file altogether. The metamodel should largely not operate so differently from other models, and including the cross product of all schema modules times formats is not a scalable strategy. 

See
https://github.com/linkml/linkml/issues/546

sneakers-the-rat:
> I realize the current SOP for these kinds of tests is suboptimal, so we can strategize ways to improve things.

We could make a [manually triggered](https://docs.github.com/en/actions/using-workflows/manually-running-a-workflow) action that runs the linkml tests along with the PR version of linkml-runtime, and make a branch protection rule that says [that action must be run and passing](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/about-protected-branches#require-status-checks-before-merging) before merging to main? that would make it so we don't have to run all the tests all the time, but since these are tightly coupled packages, ensures that changes here don't break linkml.

> But another approach might be to eliminate dependency on this file altogether. The metamodel should largely not operate so differently from other models,

I think it would be very useful to be able to at least get a reliable handle to the metamodel schema files from `linkml-runtime` - if not for perf reasons like using local schema rather than always dereferencing, then for introspection reasons. i agree that metamodel shouldn't be *that* special cased, but it still is *a little* special.

I could go either way on the github links part, i am not really sure when that would be useful.

> including the cross product of all schema modules times formats is not a scalable strategy.

true, but with caching these tests would be pretty fast: 
- the file existence tests take essentially no time
- the github API call is the slowest part, but is the same across all tests and so can be cached
- the simple `get`s are not that expensive, but still sorta expensive, but if we have a module that says it provides links, then we should ensure that they work.

so if perf is a concern, imo we should drop the link generation part.

tests/test_linkml_model/test_linkml_files.py
sneakers-the-rat: oop no

cmungall:
> I think it would be very useful to be able to at least get a reliable handle to the metamodel schema files from linkml-runtime

oh absolutely - my point was just that having an enormous list of N x M constants is not necessarily the best way to do this

sneakers-the-rat:
ah i see. yes totally agreed - ideally this metadata should come from the generators, no? or maybe projectgen? i'm actually still a little foggy on the workflow for updating the `linkml_model` module but ya for sure on the same page re: a ton of constants not being a good longterm strategy

PR #312, opened 2024-03-21
Make tests fail, not print

sneakers-the-rat:
The output of the tests is pretty difficult to read, but there were lots of places that were saying "ERROR" without failing. 

Looking further, it appears as if the default behavior for the TestEnvironment is to print, not fail on error. 

There are also many print statements throughout the tests - raising an exception should be the print statement in tests.

This PR
- Changes the default test behavior to fail on errors
- Raises exceptions rather than logging when tests should fail
- Removes all print statements from tests.

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -24,7 +24,7 @@ test_settings = config['test.settings']
 # Action on mismatch.  One of 'Ignore', 'Report' or 'Fail'
 #  If 'Fail', the expected file will be saved in the appropriate temp directory
 #  NOTE: Before setting this back to Report or Ignore, you need to run cleartemp.sh in this directory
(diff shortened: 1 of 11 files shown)


sneakers-the-rat:
The errors here are from comparisons being done to output files that don't exist - so the test is trivial, and the errors are correct, but should be conditioned on something like `--generate-snapshots`. 

So we need to either:
- remove tests or
- add snapshots

The fact that the tests are failing on windows but the check appears to be passing in the gh interface indicates that the tests as currently configured don't actually alert us to failures (for windows). I assume that's from some weirdness in the way `unittest` deals with return codes, so we should switch to pytest ASAP.

amc-corey-cox:
@sneakers-the-rat I have moved the work on this PR into the repo branch tests-fail . If you would like to continue this work please re-open the PR on the mono-repo after the merge from that branch. We'll lift it over for you so it should take less effort. This will have a slight rename in the linkml repo but you will be able to re-open it there.

PR #313, opened 2024-03-21
Add inheritance ordering, unify element ordering

sneakers-the-rat:
Schemaview seems like the right place to keep sorting methods that will be used by all the generators. Lo and behold, it already is!

Added the inheritance ordering method used by pythongen and pydanticgen, put all ordering into a single method.

Must be merged before the adjoining linkml PR (will add link in a second) and linkml lockfile must be regenerated to reflect the release containing this PR

### DIFF
--- a/linkml_runtime/utils/schemaview.py
+++ b/linkml_runtime/utils/schemaview.py
@@ -6,7 +6,7 @@ from functools import lru_cache
 from copy import copy, deepcopy
 from collections import defaultdict, deque
 from pathlib import Path
-from typing import Mapping, Tuple
+from typing import Mapping, Tuple, TypeVar
 import warnings
 
 from linkml_runtime.utils.namespaces import Namespaces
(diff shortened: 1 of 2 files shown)


linkml_runtime/utils/schemaview.py
```diff
@@ -342,6 +370,32 @@ def _order_rank(self, elements: dict):
         rank_ordered_elements.update(unranked_map)
         return rank_ordered_elements
 
+    def _order_inheritance(self, elements: DefDict) -> DefDict:
+        """
+        sort classes such that if C is a child of P then C appears after P in the list
```
sujaypatil96, line 375: Quick nit: it's not just classes, inheritance can be applied to slots and subsets as well, so this is for _elements_ generally?

sujaypatil96 (approved):
Thanks for adding the ability to order elements by inheritance @sneakers-the-rat   we might want to use this functionality in the documentation generator at some point as well!

The code changes look good to me, this PR is ready to go  🚀

linkml_runtime/utils/schemaview.py
sneakers-the-rat: Totally - I was just copying the existing comment there. Specifically the type bound is for Definitions since you need the `is_a` property

PR #314, opened 2024-03-21
pretty print models

sneakers-the-rat:
hi i'm ​🇸​​🇹​​🇷​​🇮​​🇳​​🇬​ ​🇷​​🇪​​🇵​​🇷​​🇪​​🇸​​🇪​​🇳​​🇹​​🇦​​🇹​​🇮​​🇴​​🇳​ ​🇴​​🇫​ ​🇲​​🇴​​🇩​​🇪​​🇱​ ​🇩​​🇪​​🇫​​🇮​​🇳​​🇮​​🇹​​🇮​​🇴​​🇳​

you might remember me from other print statements such as

<code>
>>> sv = SchemaView('tests/test_loaders_dumpers/input/personinfo.yaml')
>>> cls = sv.get_class('Person')
>>> print(cls)
[...]

PR #319, opened 2024-04-03
Add action to test with upstream linkml v2

sneakers-the-rat:
Continues: https://github.com/linkml/linkml-runtime/pull/316

Sorry to keep doing this with CI actions - as far as I know I can't test them on the fork until they are in main?

The previous action wouldn't run because the dynamic versioning plugin can't detect version when run within a poetry shell - https://github.com/mtkennerly/poetry-dynamic-versioning?tab=readme-ov-file#caveats

you also apparently need to explicitly install the plugin.

so i modified it so that we create an arbitrarily high version with the commit hash appended to it so we can be sure that the current version is being installed rather than a cached version.

You can see that the action is running *but correctly failing due to https://github.com/linkml/linkml/issues/2047 * here - https://github.com/sneakers-the-rat/linkml-runtime/actions/runs/8531972641/job/23372442173

this should get squash merged, but wanted to preserve commit history in PR so it's possible to see the various options i tried. blah. one day this will all be done and our CI will be spic and span

### DIFF
--- a/.github/workflows/test-upstream.yaml
+++ b/.github/workflows/test-upstream.yaml
@@ -71,7 +71,13 @@ jobs:
 
       - name: remove potentially cached linkml-runtime version
         working-directory: linkml
-        run: poetry run pip uninstall linkml-runtime
+        run: poetry run pip uninstall linkml-runtime -y
+
+      - name: get linkml-runtime short hash
+        working-directory: linkml-runtime
+        run: |
+          LINKML_RUNTIME_COMMIT=$(git rev-parse --short HEAD)
+          echo "LINKML_RUNTIME_COMMIT=$LINKML_RUNTIME_COMMIT" >> "$GITHUB_ENV"
 
       # install the local version of linkml-runtime:
       # we do this vs. pip --force-reinstall so that we keep linkml's locked deps
@@ -82,7 +88,7 @@ jobs:
       - name: add linkml-runtime to linkml's venv
         working-directory: linkml
         env:
-          POETRY_DYNAMIC_VERSIONING_OVERRIDE: "linkml-runtime = 1.99.0"
+          POETRY_DYNAMIC_VERSIONING_OVERRIDE: "linkml-runtime = 1.99.0+${{ env.LINKML_RUNTIME_COMMIT }}"
         run: poetry add ../linkml-runtime
 
       # note that we run the installation step always, even if we restore a venv,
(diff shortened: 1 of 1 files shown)


.github/workflows/test-upstream.yaml
```diff
@@ -24,6 +24,9 @@ jobs:
           - python-version: "3.10"
             pydantic-version: "1"
```
pkalita-lbl, line 25: I know it's not even germane to the changes here but I recently simplified the testing matrix in the `linkml` repo (https://github.com/linkml/linkml/pull/2041). It might be worth using the same matrix here. I wouldn't lose sleep at night if we didn't do a Pydantic v1 test in this upstream workflow.

.github/workflows/test-upstream.yaml
```diff
@@ -61,29 +64,40 @@ jobs:
...
 
+      - name: remove potentially cached linkml-runtime version
+        working-directory: linkml
+        run: poetry run pip uninstall linkml-runtime -y
+
+      - name: get linkml-runtime short hash
+        working-directory: linkml-runtime
+        run: |
+          LINKML_RUNTIME_COMMIT=$(git rev-parse --short HEAD)
+          echo "LINKML_RUNTIME_COMMIT=$LINKML_RUNTIME_COMMIT" >> "$GITHUB_ENV"
```
pkalita-lbl, line 80: I think this step can be replaced with running this in the `linkml-runtime` directory:

```
poetry self add "poetry-dynamic-versioning[plugin]"
poetry dynamic-versioning
```

Using the dynamic versioning plugin via the CLI will modify the `version` field in `pyproject.toml` according to the current Git status and leave the change in place. 

Then in the next step you should be able to just do the `poetry add ../linkml-runtime`. I think 😂. If that works that feels a little better to me than using the fake `1.99.0+x` version.

.github/workflows/test-upstream.yaml
sneakers-the-rat: ya ya good call, i wrote this before that got merged, and we were definitely thinking the same thing ;). so would b good to match.

.github/workflows/test-upstream.yaml
sneakers-the-rat: (agreed it's ugly and i hate it lol)

sneakers-the-rat:
Thinking about what we need for PRs like https://github.com/linkml/linkml-runtime/pull/314 - thats a change that also requires a change in upstream to pass ( https://github.com/linkml/linkml/pull/2032 ). Is it possible/should we add some input arguments that allow this to be run against a specific version of upstream? Like I know we can add arguments, but is there a way to make that sorta streamlined - like in the OP I put some line that links to the branch of upstream to test against.

The manual version would just be putting in username, repo name and branch, with defaults for all three, which is not too bad.

sneakers-the-rat:
Im p sure the most recent changes to the action are better than this lol

PR #320, opened 2024-04-03
Correct `linkml_files.URL_FOR` - add URL resolutiontests

sneakers-the-rat:
Continues: https://github.com/linkml/linkml-runtime/pull/310
Fix: https://github.com/linkml/linkml/issues/2047

Previously I had fixed some problems with `linkml_files` but didn't test the URL_FOR function. This is used by the ContextGenerator and RDFGenerator, and causes linkml tests and fail and breaks them.

This PR
- fixes `URL_FOR`
- switches tests to `pytest`
- Add requests_cache
- Unskip tests for URL resolution and correctness - these are ~slow-ish, but since we're using them upstream i figure we should test them.
- Skip fail-prone and unused github URL tests - I can't see where these functions are used within the library and they give us 403's constantly, we might want to remove those? i mostly care about the local path functions here, but also i don't think we should have broken stuff lying around if we can help it

edit: confirmed this does fix `test_generators/test_rdfgen.py`


cmungall (approved):

cmungall:
Merged as the action appears broken, see:

https://github.com/linkml/linkml-runtime/pull/324

cmungall:
Retrospective note: it turns out the SoT for this file is actually in linkml_model, and any changes such as this PR get wiped out by the metamodel release process. Clearly this is not desirable, and the SoT should probably switch to this repo

PR #328, opened 2024-07-12
first draft conversion of unittest syntax to pytest syntax for schema…

sierra-moxon:
…view tests.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -22,43 +22,47 @@ jobs:
     steps:
 
       #----------------------------------------------
-      #          install poetry
-      #----------------------------------------------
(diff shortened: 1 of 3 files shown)


sierra-moxon:
prereq: https://github.com/linkml/linkml-model/pull/201

sujaypatil96 (commented):
I agree with the behaviour of the tests -- 
* `test_induced_range_expression_with_range()`
* `test_induced_range_only_range()`
* `test_induced_range_only_range_expression()`

I tried to review the code but I can't say that I've understood it fully. Just a quick question, similar to `any_of` and `exactly_one_of` do we need to implement special handling for `none_of` or `all_of`?

PR #329, opened 2024-07-23
[perf] Cache schemaview hash

sneakers-the-rat:
Here's a fun one:

How come ~1/4 of the time in `induced_slot` is spent on `SchemaView.__hash__`?

Screenshot 2024-07-22 at 6 55 34 PM

In my tests generating the NWB schema, the `__hash__` method is called 1612428 times, that's a lot! by comparison `induced_slot` is only called 3004 times in these tests.

Screenshot 2024-07-22 at 6 58 41 PM

How come `__hash__` gets called so many times, or i guess a better question is why is it called at all?  Turns out it's called every time a method wrapped with `lru_cache` is called - since `self` is an argument to a method, and `lru_cache` uses the hashes of the arguments to match cached results, it has to hash `SchemaView` every time, but the value is always the same (except when `modified` is incremented).

So this PR just stores the value of the hash and returns that, invalidating when `modified` is incremented. ezpz

### DIFF
--- a/linkml_runtime/utils/schemaview.py
+++ b/linkml_runtime/utils/schemaview.py
@@ -6,7 +6,7 @@ from functools import lru_cache
 from copy import copy, deepcopy
 from collections import defaultdict, deque
 from pathlib import Path
-from typing import Mapping, Tuple, TypeVar
+from typing import Mapping, Optional, Tuple, TypeVar
 import warnings
 
 from linkml_runtime.utils.namespaces import Namespaces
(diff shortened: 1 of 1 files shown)


cmungall (approved):

sneakers-the-rat:
running into this again, and in case we were holding off bc the X is red, this was before we fixed the upstream tests (but i still can't trigger them, so manually trigger to check and we can get things a lil faster for free :)

edit: just rebased and now the check is green, but i would still want to see an upstream test before we merge to be sure :)

PR #335, opened 2024-07-31
Modify `induced_class()` / `induced_slot()` so it materializes non-scalar metaslots as well

sujaypatil96:
Fixes https://github.com/linkml/linkml/issues/2224

There are 3 ways in which we could have written https://github.com/linkml/linkml-runtime/blob/798591b29661efa8fef05703c79e3614d40f911d/linkml_runtime/utils/schemaview.py#L1376

1. `if v2:`
2. `if v2 is not None and ((isinstance(v2, (dict, list)) and v2) or (isinstance(v2, JsonObj) and as_dict(v2))):`
3. `if not is_empty(v2):`

Definition of `is_empty()`
https://github.com/linkml/linkml-runtime/blob/798591b29661efa8fef05703c79e3614d40f911d/linkml_runtime/utils/formatutils.py#L112-L121


cmungall (approved):

sneakers-the-rat:
well it looks like it made the upstream tests fail in an interesting way. will put my comments on the actual PR in a review but first...

the error is here, keyerror on `domain_of`: https://github.com/linkml/linkml/blob/c26d8995b164223d0dc074721fc30cf96e4965e0/tests/test_generators/test_docgen.py#L320

first, the test is wrong - the source schema does not have `domain_of` set for `Person.slot_usage.species_name` (nor would it really make sense to set `domain_of` in `slot_usage`): https://github.com/linkml/linkml/blob/c26d8995b164223d0dc074721fc30cf96e4965e0/tests/test_generators/input/kitchen_sink.yaml#L142-L143

what's sort of funny is that what must have been happening there is that the list for `Person.slot_usage['species_name'].domain_of` was bound to `Person.species_name.domain_of` since that's what the old behavior was, so then later when that list is populated then the `slot_usage` version of that was also populated since it's the same object. 

`slot` gotten earlier in the method is shallow copied so that should break the object identity relation, but `ancestor_class.slot.slot_usage` is *not*, so it is vulnerable to mutation, hence why even with the old old behavior of multiple deepcopies in `induced_slot` that this behavior persisted in the test. 

that would be a *real big problem* if nonscalar slots metaslots were ever mutated outside of the schemaview (unfortunately they are), since then *any values* set *on the slot* would have the wildly unpredictable behavior of propagating back up to the *slot usage* of the last ancestor to have it set, which would then *propagate back down* to any other slots that inherit from it.

this is additionally a problem because `induced_slot` is often called twice or three times for each slot despite caching, depending on the generator (for each time the signature differs), so we should expect this to lead to nondeterministic behavior that depends on the number of times a slot is generated and how its inheritance tree branches * how slot_usage is used in it.

so anyway, obvi this PR should only be merged along with a patch to the upstream tests, but i think we are probably in for a bit of a ride as we figure out what other unanticipated behavior this method has in store for us lol

tests/test_issues/test_issue_2224.py
```diff
@@ -0,0 +1,52 @@
...
+        # check that the annotations (dictionary) is being materialized properly
+        self.assertIsInstance(cls.attributes["jog_wheels"].annotations, JsonObj)
+        self.assertEqual(
+            cls.attributes["jog_wheels"].annotations.expected_value.value,
+            "an integer between 0 and 4",
+        )
+        self.assertEqual(
+            cls.attributes["volume_faders"].annotations.expected_value.value,
+            "an integer between 0 and 8",
+        )
```
sneakers-the-rat, line 36: i appreciate the whimsicality of this test, but if it were up to me i would keep the unit test constrained to the behavior in question :) these values should be unaffected by the bug (since DJController only overrides `tempo` with `slot_usage`)

tests/test_issues/test_issue_2224.py
```diff
@@ -0,0 +1,52 @@
...
+
+from tests.test_issues.environment import env
+
+
+class Issue2224TestCase(TestCase):
+    env = env
+
+    def test_issue_2224_slot_classes(self):
+        sv = SchemaView(env.input_path("linkml_issue_2224.yaml"))
+        cls = sv.induced_class("DJController")
```
sneakers-the-rat, line 14: again this is just a 'if it were up to me thing', but i find organizing tests by issue number to be a bad pattern - it's hard to know what this test is testing for from the test results, and then when evaluating "the tests for schemaview" it's hard to know what tests already exist and what don't, so we end up with multiple tests for the same thing. I would call this something like "test_induced_slot_usage_overwrite" or whatever and then put a description and link to the issue in the docstring so that when the test fails we can be like "oh something is wrong with the way that induced slots are overwriting slot usage," and when we need to test similar things we can ctrl+f around to find where we should put the next test near.

tests/test_issues/input/linkml_issue_2224.yaml
```diff
@@ -0,0 +1,62 @@
...
+    slot_usage:
+      tempo:
+        examples:
+        - value: 120.0
+        - value: 144.0
+        - value: 126.8
+        - value: 102.6
+        annotations:
+          expected_value: a number between 0 and 300
+          preferred_unit: BPM
```
sneakers-the-rat, line 22: only override one of `examples` or `annotations` here, test that the non-overridden one has its values propagated from the original slot definition

tests/test_issues/test_issue_2224.py
```diff
@@ -0,0 +1,52 @@
+import unittest
+from unittest import TestCase
+from linkml_runtime.utils.schemaview import SchemaView
+from jsonasobj2 import JsonObj
+
+from tests.test_issues.environment import env
+
+
+class Issue2224TestCase(TestCase):
```
sneakers-the-rat, line 9: CI already runs pytest, and if the goal longterm is to move to pytest, might be a good idea to write things as pytest tests from here on out :)

tests/test_issues/test_issue_2224.py
```diff
@@ -0,0 +1,52 @@
...
+            cls.attributes["tempo"].annotations.expected_value.value,
+            "a number between 0 and 300",
+        )
+        self.assertEqual(
+            cls.attributes["tempo"].annotations.preferred_unit.value, "BPM"
+        )
+
+
+if __name__ == "__main__":
+    unittest.main()
```
sneakers-the-rat, line 52: to run single tests in pytest: `pytest -k "test_name_of_test"`

linkml_runtime/utils/schemaview.py
```diff
@@ -1369,7 +1370,10 @@ def induced_slot(self, slot_name: SLOT_NAME, class_name: CLASS_NAME = None, impo
                         if v2 is not None:
                             v = COMBINE[metaslot_name](v, v2)
                     else:
-                        if v2 is not None:
+                        # can rewrite below as:
+                        # 1. if v2:
+                        # 2. if v2 is not None and ((isinstance(v2, (dict, list)) and v2) or (isinstance(v2, JsonObj) and as_dict(v2)))
```
sneakers-the-rat, line 1375: if we ever turn on linting for this repo this line is 111 chars long, so if we linebreak now we avoid one in the future <3

tests/test_issues/test_issue_2224.py
```diff
@@ -0,0 +1,52 @@
...
+            "an integer between 0 and 4",
+        )
+        self.assertEqual(
+            cls.attributes["volume_faders"].annotations.expected_value.value,
+            "an integer between 0 and 8",
+        )
+
+        # examples being overriden by slot_usage modification
+        for example in cls.attributes["tempo"].examples:
+            self.assertIn(example.value, ["120.0", "144.0", "126.8", "102.6"])
```
sneakers-the-rat, line 40: if the slot value *wasn't* being overridden (ie. it was still `120` and `144`, this would still pass. test for exact values

(after adding `from linkml_runtime.linkml_model import Example` above)

```suggestion
        assert cls.attributes["tempo"].examples == [Example(value='120.0'), Example(value='144.0'), Example(value='126.8'), Example(value='102.6')]
```

sneakers-the-rat (changes requested):
change to `induced_slot` looks fine! probably the best we can do given the state of `JsonObj` et al making stuff so expensive. It does add a few %s more runtime per call to what is already the most expensive method in the library, which is not great, but it is what it is.

reason why i'm requesting changes is that the test here doesn't test for the problem in https://github.com/linkml/linkml/issues/2224 - this test passes when I run it against the old behavior, while the [MWE here](https://github.com/linkml/linkml/issues/2224#issuecomment-2249214252) correctly fails. good practice when bugfixing - write the test first, check that it fails, fix the bug, check that tests pass.

the bug is that slot usage for an unrelated metaslot causes a different nonscalar metaslot to be overridden - ie. `tempo` should have some nonscalar metaslot value set, `DJController` should have `slot_usage` for a *different* metaslot set, and then the test should assert that the first metaslot is unchanged. So, only override one or the other attr, check that the non-overridden one is the value of the base definition, check that the overridden one is the value that's set in `slot_usage`.

rest of comments are style things intended for maintainability :)

tests/test_issues/test_issue_2224.py
sneakers-the-rat: Or make the parent and overridden slot have a single value and change that value, dats one of the reasons to keep unit tests as minimal as possible so then you also can be safely lazy (complimentary, non-derogatory) in the asserts :)

linkml_runtime/utils/schemaview.py
sujaypatil96: Good suggestion! I'll break this line up/structure the comment such that it is black / ruff rules compliant.

tests/test_issues/input/linkml_issue_2224.yaml
sujaypatil96: Makes sense! i'll remove the overrides for `annotations` (retaining the overrides for `examples` only).

Will check that the original `annotations` from `tempo` are being propagated as expected:

```yaml
    annotations:
      expected_value: a number between 0 and 200
      preferred_unit: BPM
```

tests/test_issues/test_issue_2224.py
sujaypatil96: Makes complete sense, and good suggestion again! I'll move this test into `tests/test_utils/test_schemaview.py`.

tests/test_utils/test_schemaview.py
```diff
@@ -945,6 +947,34 @@ def test_is_inlined(self):
                 actual_result = sv.is_inlined(slot)
                 self.assertEqual(actual_result, expected_result)
 
+    def test_materialize_nonscalar_slot_usage(self):
```
sneakers-the-rat, line 950:
```suggestion
    def test_materialize_nonscalar_slot_usage(self):
        """
        ``slot_usage`` overrides values in the base slot definition without
        clobbering unrelated, nonscalar values.
        
        See: 
        - https://github.com/linkml/linkml/issues/2224
        - https://github.com/linkml/linkml-runtime/pull/335
        """
```

Just giving us a little hint for our future selves that can be skimmed :) am on phone, sorry if this is broken.

sneakers-the-rat (approved):
Looks great! Only thing I would change is adding a docstring to the test asserting what it tests, but good with me with or without ♥

tests/test_utils/test_schemaview.py
sneakers-the-rat: Something about that docstring is not working with the code highlighting, thats what I get for reviewing from a phone lol

sneakers-the-rat:
Before we merge we should re-run upstream tests and patch that at the same time so we dont forget :)

sujaypatil96:
I looked at the upstream linkml test that's causing the error in [test_docgen.py](https://github.com/linkml/linkml/blob/main/tests/test_generators/test_docgen.py), and I think it's okay to remove this assertion: https://github.com/linkml/linkml/blob/main/tests/test_generators/test_docgen.py#L320-L321

The contents of `person_dict["slot_usage"]` looks like below, there's no `domain_of` appearing in the materialization anymore:

```python
{
    "name": {"name": "name", "pattern": "^\\S+ \\S+$"},
    "species name": {"equals_string": "human", "name": "species name"},
    "stomach count": {"equals_number": 1, "name": "stomach count"},
}
```

sneakers-the-rat:
Actually yno what would be a good idea is writing a specific test for that *not* happening. I can follow up with that

sujaypatil96:
> Actually yno what would be a good idea is writing a specific test for that not happening. I can follow up with that.

Thank you!! will you make a PR for this? i'll be happy to review it 😁 once that patch is ready to go, we can merge this PR in first, and then the patch PR on linkml.

sneakers-the-rat:
Lets put that test over here, no? Since its a test of schemaview behavior?

cmungall:
@sierra-moxon noticed that the example here: https://github.com/linkml/linkml/issues/2224#issuecomment-2249214252 has an override, can we change the test example so the range at the slot level uses `Any` as a range?

sujaypatil96:
Made the change requested [here](https://github.com/linkml/linkml-runtime/pull/335#issuecomment-2264466010) in this commit ([ba677c6](https://github.com/linkml/linkml-runtime/pull/335/commits/ba677c6fb8576390cf6a508ed49e0329cd17dc48))

PR #336, opened 2024-08-03
Handle attributes edge case in schema_as_dict

bob-chevalier:
When using SchemaBuilder, any slots defined with name="name", will be removed by the schema_as_dict function, if the slot was added to a class using the use_attributes=True flag.

See issue: https://github.com/linkml/linkml/issues/2247


cmungall (approved):

bob-chevalier:
Hi @cmungall. Thanks for the approval.  It looks like 3 of the checks failed while Poetry attempted to install a dependency.  My PR doesn't touch the lock file so I was hoping this might be a transient failure, but I don't see a way to re-trigger the checks.

cmungall (approved):

PR #338, opened 2024-08-27
BF: Fix bug and other improvement in `SchemaBuilder.add_class()`

candleindark:
This PR does the following.

1. Prevent the use of `kwargs` in setting the class to be added with an unacceptable attribute if `cls` is given as a `ClassDefinition` object (test provided)
2. Correct detection of duplicating classes. (test provided)
3. Handle the case `slots` is `None` while `use_attributes` is `True` (test provided)
4. Update the example in the doc string of `SchemaBuilder` to have the correct import statement (no test needed)
5. Provide documentation of the `use_attributes` param in the docstring (no test needed)
6. Give more informative documentation for the `slot_usage` param according to the behavior of existing code. (no test needed)
7. Provide type annotation to the `replace_if_present` and `use_attributes` params (no test needed)

Point 1 is included in this PR because the following lines are unneeded for the cases in which `cls` is given as a `dict` or `str` and are inconsistent with the behavior of `SchemaBuilder.add_slot()` and `SchemaBuilder.add_enum()`. In both `SchemaBuilder.add_slot()` and `SchemaBuilder.add_enum()`, if the first argument is of the target type, `SlotDefinition` and `EnumDefinition` respectively, `kwargs` is ignored.

https://github.com/linkml/linkml-runtime/blob/9fd5a3e374a1d1820da07a0e3d165b8f667033f4/linkml_runtime/utils/schema_builder.py#L101-L102

### DIFF
--- a/linkml_runtime/utils/schema_builder.py
+++ b/linkml_runtime/utils/schema_builder.py
@@ -66,6 +66,7 @@ class SchemaBuilder:
 
         :param cls: name, dict object, or ClassDefinition object to add
         :param slots: slot of slot names or slot objects.
+            Note: If `use_attributes=True`, then this must be a list of SlotDefinitions
         :param slot_usage: slots keyed by slot name
         :param replace_if_present: if True, replace existing class if present
         :param kwargs: additional ClassDefinition properties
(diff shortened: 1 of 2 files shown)


yarikoptic:
If you could add a test which would demonstrate (fail) the issue without this fix, that would be great.

cmungall (changes requested):
Thanks @candleindark - do you think you can add the test as @yarikoptic suggests?

candleindark:
> Thanks @candleindark - do you think you can add the test as @yarikoptic suggests?

I think I can do that. Any suggestion in where is the best place to put these tests.

cmungall:
Hmm, I thought we had a test_schemabuilder here https://github.com/linkml/linkml-runtime/tree/main/tests/test_utils but it seems not... if you can start one that would be great. We favor simple pytest style tests.

cmungall (approved):

PR #343, opened 2024-10-02
convert schemaview unittests to pytest

sierra-moxon:


tests/test_utils/test_schemaview.py
```diff
@@ -34,998 +31,872 @@
...
+@pytest.fixture
+def view():
+    """Fixture for SchemaView with imports."""
+    return SchemaView(SCHEMA_WITH_IMPORTS)
 
-class SchemaViewTestCase(unittest.TestCase):
 
-    def test_children_method(self):
+def test_children_method():
         view = SchemaView(SCHEMA_NO_IMPORTS)
```
pkalita-lbl, line 46: You don't need to redefine `view` in all these test functions since you made a fixture for it. Just include `view` or `schema_view_no_imports` as an argument to the function.

pkalita-lbl (commented):
Overall looks great!

I'd just double check that `make test` isn't running the other unconverted tests twice.

Makefile
```diff
@@ -11,6 +11,8 @@ update_model:
 
 test:
 	poetry run python -m unittest discover
+	poetry run pytest
```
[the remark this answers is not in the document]
sierra-moxon: done.

tests/test_utils/test_schemaview.py
sierra-moxon: yep, makes sense.  Some of the tests redefine the import to test components of importing but not this.  Fixed (I replaced the new SchemaView initialization with the fixture but left the `view` variable assignment in place because not doing so would mean many more test changes.)

pkalita-lbl (approved):

PR #346, opened 2024-10-25
fix(permissiblevalue): serialization

Silvanoc:
Serialization of `PermissibleValue(text='value')` should be `'value'`, instead it is `PermissibleValue({"text": 'value'})`. This patch fixes it.

Fixes https://github.com/linkml/linkml/issues/2382


Silvanoc:
This code is generated, so merging this does not make any sense. Therefore closing it. The generator will have to be fixed.

PR #348, opened 2024-11-09
Fix: rdflib dumper ignores explicitly defined prefixes

sneakers-the-rat:
Related to: https://github.com/mapping-commons/sssom-py/pull/558

copy and pasting discussion from slack:

- first is that schemaview assigns from the default_curi_maps after explicitly defined prefixes - so those prefixes in the default curi map (which is the source of the allcaps HSAPDV) override explicit prefixes. explicit assignment should always override defaults
- second is in the Namespaces object. currently it rejects repeated assignment to the same key. since it is a case insensitive dict, HsapDv is the same key as HASPDV . That shouldn’t be the case - explicit assignment should always override existing values
- and then finally, as noted above, the rdflib dumper makes repeated calls to schemaview.namespaces() and tries to assign to it, which does nothing (or should do nothing, and if it does something then it’s a bug and a side effect of mutating values in the cache which 100% should not be relied on). So i instead made a single call to schemaview.namespaces(), assigning that to a variable, and then mutated that variable instead in the as_rdf_graph method.

opening this as a draft because i don't know this code, it fixes the linked issue, but i don't have a good sense of why things are the way they are currently.

### DIFF
--- a/linkml_runtime/dumpers/rdflib_dumper.py
+++ b/linkml_runtime/dumpers/rdflib_dumper.py
@@ -47,26 +47,26 @@ class RDFLibDumper(Dumper):
             # TODO replace with `prefix_map = prefix_map.bimap` after making minimum requirement on python 3.8
             prefix_map = {record.prefix: record.uri_prefix for record in prefix_map.records}
         logger.debug(f'PREFIXMAP={prefix_map}')
+        namespaces = schemaview.namespaces()
         if prefix_map:
             for k, v in prefix_map.items():
                 if k == "@base":
-                    schemaview.namespaces()._base = v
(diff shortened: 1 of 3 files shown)


matentzn:
I just want to bump the importance of this PR for me as we make extensive use of RDFDumper in SSSOM py and right now, the outputs don't respect the provided prefix maps..

cmungall (approved):

ticapix:
could we get it merged ?

sneakers-the-rat:
Just as a note - this PR didnt have tests with it so would encourage ppl who are affected by this bug to follow on with tests to ensure against regression

PR #349, opened 2024-11-18
Remove import of re submodule from typing

caufieldjh:
See https://github.com/linkml/linkml/issues/2420


caufieldjh:
oh, this is redundant vs. #345, and that does a much more thorough job of updating for 3.13

PR #353, opened 2024-12-20
Fix python 3.13 deprecation of typing.re

nicholsn:
In python 3.13, `typing.re` was removed (deprecated since Python 3.8). See https://github.com/python/cpython/issues/92871

This pr refactors to support python 3.13 which currently throws

```python
ImportError: cannot import name 're' from 'typing' 
```


nicholsn:
Ok, I think that last commit should work - @sierra-moxon can you trigger the CI?

nicholsn:
Ya I realized this was way more involved once I dug in. I'll close the PR

PR #356, opened 2025-01-01
feat: add language and datatype support when converting rdflib.terms.…

ticapix:
Proposal to address https://github.com/linkml/linkml/issues/2475

Without this, it's impossible to convert a Python Instance of a LinkML gen-python class to RDF without loosing language and datatype information.


sierra-moxon:
@turbomam @ptgolden  - if you have a moment, can you please review this RDFlib linkml PR?

amc-corey-cox:
@ticapix I have moved the work for this PR from the contributor fork to the linkml-runtime branch https://github.com/linkml/linkml-runtime/pull/new/add_rdf_lang_type_support  . These branches will be lifted over in the merge to the mono-repo so you won't lose you work. You will need to open a new PR against the linkml monorepo after the move. I'd prefer to get this in but if we can't please open a new PR when you resume work. Thanks!

tests/test_issues/test_linkml_runtime_issue_2475.py
ptgolden: I'm not sure about the strategy here to define a schema inline-- if this test were merged, the definition would potentially have to be kept up to date with changes in behavior of linkml runtime. (There's already one here, as `dataclasses_init_fn_with_kwargs` has been deprecated).

There's already a `Person` schema defined in tests. Can that be reused here to define a test case? (See [these lines](https://github.com/linkml/linkml-runtime/blob/9e0e62b6270ba9475f5d3a2e04d7d50f560e9af0/tests/test_loaders_dumpers/test_rdflib_dumper.py#L16C1-L32)).

It might be good to add a new test case to that file instead of adding a new file here. (Although others may have a different opinion). There's another test there that is doing something very similar to what this test is doing:

```
# see https://github.com/linkml/linkml/issues/429
SCHEMA_429 = INPUT_PATH / "personinfo_test_issue_429.yaml"
DATA_429 = INPUT_PATH / "example_personinfo_test_issue_429_data.yaml"
OUT_429 = OUTPUT_PATH / "example_personinfo_test_issue_429_data.ttl"

...

@pytest.fixture
def issue_429_graph():
    """Create RDF graph for issue 429 testing."""
    view = SchemaView(str(SCHEMA_429))
    container = yaml_loader.load(str(DATA_429), target_class=Container_429)
    rdflib_dumper.dump(container, schemaview=view, to_file=str(OUT_429))
    g = Graph()
    g.parse(str(OUT_429), format="ttl")
    return g
```

PR #357, opened 2025-01-09
Support isodate ending with Z

dalito:
The PR switches (for Python <3.11) to [isodate](https://pypi.python.org/pypi/isodate/) package for isodate parsing. It is already installed with LinkML since it is a rdflib dependency.

Todo (maybe -> done!):

- [x] restrict the use of the `isodate` package to Python < 3.11. Starting with Python 3.11 the standard library´s `datetime.datetime.fromisoformat` is able parse isodates ending with "Z" correctly.

Closes linkml/linkml#2096
Closes linkml/linkml#629

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -18,7 +18,7 @@ jobs:
           python-version: "3.8"
 
     runs-on: ${{ matrix.os }}
-
+    
     steps:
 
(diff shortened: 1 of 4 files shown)


dalito:
I cleaned up the commit history to make reviewing easier.

cmungall (approved):

PR #358, opened 2025-01-10
Fail actions on error correctly on Windows

dalito:
I also made this change in #357 but wanted to see the Windows test results without any other change.

On Windows failing tests were not correctly detected. The run was still "green" even if errors occured. See https://github.com/actions/runner-images/issues/6668 for more.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -49,6 +49,8 @@ jobs:
       #              coverage report   
       #----------------------------------------------
       - name: Generate coverage results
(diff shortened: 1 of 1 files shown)


dalito:
Unfortunately, some Windows errors slipped in undetected over time.

For the main LinkML repo I fixed the same problem already 2022 but obviously forgot about other sub-repos like this one 😲

dalito:
This PR is replaced by #360 which also adds the required fixes.

PR #359, opened 2025-01-11
Fix hidden windows failures

dalito:
This fixes two Windows path-related issues that were undetected because of the gh-action.

See runs in #358 for the hidden failures that are fixed by this PR.

Closes linkml/linkml#2490

### DIFF
--- a/linkml_runtime/linkml_model/linkml_files.py
+++ b/linkml_runtime/linkml_model/linkml_files.py
@@ -1,4 +1,4 @@
-from pathlib import Path
+from pathlib import Path, PurePath
 from enum import Enum, auto
 from typing import Dict, Optional, Union, Tuple, NamedTuple
 from urllib.parse import urljoin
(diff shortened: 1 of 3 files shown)


dalito:
Closed because of wrong issue number in branch name.

linkml_runtime/linkml_model/linkml_files.py
```diff
@@ -152,7 +152,7 @@ def LOCAL_PATH_FOR(source: Source, fmt: Format) -> str:
 
 
 def GITHUB_IO_PATH_FOR(source: Source, fmt: Format, version="latest") -> str:
-    path = '/'.join([version, 'linkml_model', *_build_path(source, fmt)])
```
sneakers-the-rat, line 155: This is for a URL, not an OS path, which is why it joins on `/` rather than using pathlib. Looks like the pathlib impl should be equivalent but just explaining the manual string munging

PR #362, opened 2025-01-20
Fix SchemaView metadata injection and get URI

vincentkelleher:
I noticed an issue while working on our Gaia-X ontology as there is a slot named `uri` and also a type (from LinkML types) named `uri`. This caused the metadata injection (overwriting `from_schema` with the current schema ID) to fail as `dict`s are used and merged, meaning either the LinkML `uri` could get the correct schema ID or the Gaia-X `uri` slot not both.

To fix this I just proposed to merge the values in a list avoiding any collision.

Please note I also fixed the `get_uri(..)` method as in the case above I would get a `StopIteration` exception coming from the `next(..)` method arriving at the end of the schema map without finding element. To do so, I just added a default value to the iteration method: `next(<generator function>, <fallback/default value>)`.

### DIFF
--- a/linkml_runtime/utils/schemaview.py
+++ b/linkml_runtime/utils/schemaview.py
@@ -318,7 +318,16 @@ class SchemaView(object):
 
         if inject_metadata:
             for s in self.schema_map.values():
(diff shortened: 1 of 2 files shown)


amc-corey-cox (approved):
Okay, this looks good to me!

PR #363, opened 2025-01-28
Poetry 2 compatibility

dalito:
This only prepares for Poetry 2.0.x by removing old pins and changing the syntax to pass `poetry check` without warning for poetry 1.8.5.

With poetry 2.0.x there are warnings but no errors. So with the PR merged the project can be installed just fine with poetry 2.0.x.

The poetry 2.0 update of `pyproject.toml` will change more and should maybe done in another PR. Moreover, we may want to wait a little before breaking Poetry 1.8.x compatibility.


.github/workflows/test-upstream.yaml
```diff
@@ -53,7 +53,9 @@ jobs:
           python-version: ${{ matrix.python-version }}
 
       - name: Install poetry
-        run: pipx install poetry
+        # Pin to 1.8.5. because Poetry 2.0.0 installs plugins in a different way
+        # (dynamic versioning in next step)
+        run: pipx install poetry==1.8.5
```
sneakers-the-rat, line 58: It looks like there is a way to pin a specific version of poetry in pyproject.toml https://python-poetry.org/blog/announcing-poetry-2.0.0/

We may want to do a quick 1-line PR to upstream linkml to add this pin there rather than here so they don't go out of sync & fix later, but since this is likely to be a quick and easy change to both packages not necessary at all and this would be fine too

sneakers-the-rat (approved):
I am not seeing the pass `poetry check` with a warning part, but if this patch allows CI to pass pending a full migration then looks good to me

.github/workflows/test-upstream.yaml
dalito: I believe the option to[ constrain poetry](https://python-poetry.org/docs/pyproject#requires-poetry) was added only in 2.0 - I could not find it in 1.8.5.

dalito:
I now think it is better to directly go to Poetry 2.0 fully.

Yesterday, I saw a failure locally which made me think that this fix is required but testing with a fresh checkout of the current main-branch shows that `poetry check`, `poetry lock` and `poetry install` work with 1.8.5 and 2.0.1 with warning only.

sneakers-the-rat:
sounds good to me :) hopefully not too complicated, the 2.0 format is the standard anyway

PR #360, opened 2025-01-11
Fix hidden windows failures

dalito:
This fixes two Windows path-related issues that were undetected because of the gh-action.

See runs in #358 for the hidden failures that are fixed by this PR.

Closes linkml/linkml#2490

PR #364, opened 2025-01-29
Poetry 2 upgrade (PEP612-conformity) and gh-actions renovation

dalito:
- change `pyproject.toml` to PEP612 style for poetry 2.x
- update all gh-actions for poetry 2.x
- change action to load other actions by hash-ID (the PR updates each to its latest release)
- enable dependabot for monthly action-updates (will create a PR not autoupdate)
- prepared `pypi-publish` action for [trusted publishing](); it is commented out now since you need to change the configuration in PyPI as well to enable it. Note that the env name is not `pypi`as in the screenshots of the PyPI-docs but `pypi-release`.
- simplified the action test-upstream by using poetry sync
- figured out how to install poetry-dynamic-versioning that it actually works in gh-actions with current poetry 2.0.1

Ignore the windows test failures or merge #360 first. 😄 

Closes linkml/linkml#2520


.github/workflows/check-dependencies.yaml
```diff
@@ -26,17 +26,16 @@ jobs:
...
-        # Pin to 1.3.2 to workaround https://github.com/python-poetry/poetry/issues/7611
-        run: pipx install poetry==1.3.2
+        run: pipx install poetry
 
       #----------------------------------------------
       #       check-out repo and set-up python
       #----------------------------------------------
       - name: Check out repository
-        uses: actions/checkout@v3
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
```
sneakers-the-rat, line 35: I have found using exact hash pins to be a bit of a pain without much benefit, but no strong objection

.github/workflows/check-dependencies.yaml
sneakers-the-rat, line 1: Huh, hadn't looked at this action before. I am not exactly sure what the point is, because unless we are looking at the results of this run regularly it doesn't rly force us to do anything and isn't visible e.g. on PRs? This would probably be easier to do by just running `pip install` rather than poetry if the intention is to test against the deps people would get if they installed normally.

Just passing thoughts, not needed to change in this PR.

pyproject.toml
```diff
@@ -1,23 +1,23 @@
...
+  {name = "Chris Mungall", email = "cjmungall@lbl.gov"},
+  {name = "Harold Solbrig", email = "solbrig@jhu.edu"},
+  {name = "Sierra Moxon", email = "smoxon@lbl.gov"},
+  {name = "Bill Duncan", email = "wdduncan@gmail.com"},
+  {name = "Harshad Hegde", email = "hhegde@lbl.gov"},
 ]
-
-readme = "README.md"
-
+license = {text = "LICENSE"}
```
sneakers-the-rat, line 15:
```suggestion
license = {file = "LICENSE"}
```

And this would actually be better:

```
license = "CC0-1.0"
```

Wait why is this licensed as CC0?

pyproject.toml
```diff
@@ -1,23 +1,23 @@
...
 homepage = "https://github.com/linkml/linkml-runtime"
 repository = "https://github.com/linkml/linkml-runtime"
 documentation = "https://github.com/linkml/linkml-runtime"
-
+readme = "README.md"
```
sneakers-the-rat, line 19:
```suggestion
readme = { file = "README.md", content-type = "text/markdown" }
```

pyproject.toml
```diff
@@ -32,40 +32,44 @@ classifiers = [
     "Programming Language :: Python :: 3.11",
     "Programming Language :: Python :: 3.12",
 ]
-
-packages = [
-    { include = "linkml_runtime" }
+include = ["linkml_runtime"]
+requires-python = ">=3.8"
+dynamic = ["version", "readme"]
```
sneakers-the-rat, line 37:
```suggestion
dynamic = ["version"]
```

Readme isn't dynamic, right? Or is this some poetry thing

sneakers-the-rat (approved):
pretty straightforward.

ideally the changes to the actions would be in a separate PR so they don't hold up the changes to pyproject.toml (idk what needs to get done there to get the CD working), but this looks fine aside from two minor changes in pyproject

.github/workflows/check-dependencies.yaml
dalito: Dependabot makes PRs . It updates hashes so using hashes looses its pain. See here for an [example PR](https://github.com/nfdi4cat/voc4cat-template/pull/69). If desired it can automerge or assign to someone; I prefer to merge myself after short review of changes (which are also conveniently linked in the PRs).
The dependabot config in this PR only look for action-updates but not for python-package updates (I am sure if you though it would touch Python packages because you mentioned `pip`).

pyproject.toml
dalito: I looked [it up](https://packaging.python.org/en/latest/specifications/pyproject-toml/#readme): *If the file path ends in a case-insensitive .md suffix, then tools MUST assume the content-type is text/markdown.*
So we don't need to change.

.github/workflows/check-dependencies.yaml
dalito: Sorry, I though your comment was on dependabot, now I see its on `check-dependencies.yaml`.  - I just updated it. I don't know why its present in the repo.

pyproject.toml
sneakers-the-rat: oh that's a nice little touch :)

sneakers-the-rat (approved):

sneakers-the-rat:
what da heck is windows doing lol https://github.com/linkml/linkml-runtime/actions/runs/13048917327/job/36404669816?pr=364

dalito:
Oh. I have never seen something like that locally on Win10/11.

But I vaguely remember reading somewhere that the windows runner uses a c: and d: drive somehow. ...will have a look.

Will also look at the unix failure. Earlier there was no problem, but maybe the job got cancelled before running due to the default use of [fail-fast](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast) and the (expected) windows failure.

dalito:
The problem on Windows is cause by poetry 2.0.1 using relpath for one dir on C: and the other on d: (which obviously cannot work). There is already an issue for it https://github.com/python-poetry/poetry/issues/10028

dalito:
The problem on Windows is cause by poetry 2.0.1 using os.relpath for one dir on C: and the other on d: (which obviously cannot work). There is already an issue for it https://github.com/python-poetry/poetry/issues/10028

dalito:
Solved the problem for poetry 2 by using `pipx inject` to install dynamic-versioning plugin into the pipx-environment. The issue only occurs if the (new in poetry 2) auto-installation of plugins is used.

So now it is ready for review. The windows issues are as expected until #360 is merged.

sneakers-the-rat (approved):
after loving it while it was the only tool of its kind, i have grown to really not like poetry - this looks like it was a lot more work than it should have been, it always is. other packaging tools make lockfiles without all the drama. glad poetry finally relented and implemented it, and thanks for doing that work <3

dalito:
For the failures of `test_upstream` I created an issue linkml/linkml#2520.

dalito:
On ubuntu-latest the poetry-dynamic-versioning plugin also did not work when auto-installed, see version `0.0.0` in [pick a2eed4c remove tag listing](https://github.com/linkml/linkml-runtime/actions/runs/13082441287/job/36508542174?pr=364#step:6:32)

So I changed the actions to use `pipx inject` to install the plugin on all platforms. I also slightly simplified the action `test-upstream` by using `poetry sync`. On my work I manually triggered the workflow and [it passed](https://github.com/dalito/linkml-runtime/actions/runs/13082837813) (except on 3.8 which should not matter much).

amc-corey-cox (approved):
This also looks good to me. I appreciate your cleaning things up a little here as well. What is our strategy for fixing the windows build failures?

dalito:
> What is our strategy for fixing the windows build failures?

Merging #360? 😉 - It just needs another review.

sneakers-the-rat:
idk how pipx works but i'm noticing it's getting the wrong python version here - https://github.com/linkml/linkml-runtime/actions/runs/13116111241/job/36590689408?pr=364#step:6:16

maybe because `poetry>=2` is not available for python 3.8? seems like a problem that we can jump to a different python without knowing it tho.

dalito:
It does not matter which version is used by pipx. It is isolated from the Python installed for the workflows.

dalito:
I suggest to ignore this PR until #345 is merged. Afterwards I can either rebase and resolve conflicts or (probably better for review) split it up into smaller PRs.

dalito:
Replaced by #387

PR #365, opened 2025-01-31
Adding option to eval to prevent distribution over arrays and dicts.

cmungall:
Previously the eval function in eval_utils auto-distributes over
many-valued options. This allows for this to be controlled.

For example, in this case when looking up an object, we want to avoid
distribution:

+    pd = dict(name='x', aliases=['a', 'b', 'c'], address=Address(street='1 x street'))
+    assert eval_expr('p.name', p=pd, _distribute=False) == 'x'


amc-corey-cox (approved):
This looks good to me!

PR #366, opened 2025-02-01
Always run upstream test action

sneakers-the-rat:
when i wrote the upstream test action, we limited it to being run after approval from maintainers because the upstream tests used to take an eternity. they don't anymore, and it turns out to be a bigger headache than expected to have to trigger the tests to run every time. so this PR just makes them run like a normal PR action


dalito (approved):
I think this makes a lot of sense. PR authors and reviewers should be able to recognize problems before approval without needing to manually trigger a workflow.

PR #368, opened 2025-02-09
Fix multilayer relative imports

will-0:
Originally mentioned in [this issue](https://github.com/linkml/linkml/issues/2525) on the [linkml](https://github.com/linkml/linkml) repo.

When you attempt to use relative imports from a file that, itself, has been imported with a relative import, path resolution fails. This bug occurs both in `linkml` and `linkml-runtime`. In `linkml-runtime`, the problem occurs in `schemaview.py`:

```python
if sn.startswith('.') and ':' not in i:
       i = os.path.normpath(str(Path(sn).parent / i))
```

If both `sn` and `i` are relative file paths, the path added to `todo` is an absolute path (at least on Mac). When `i` later than becomes `sn`, it's not picked up by the above clause and anything it imports does not hit the above condition and therefore just defined relative to the base file.

Marking as draft. @sneakers-the-rat noticed you were the last to edit this section, specifically [here](https://github.com/will-0/linkml-runtime/commit/af5a1a6e40ac24f8bd0eabe6ab37d04f1021a326). Would appreciate input as to whether there's a reason for the above / it is expected behaviour and I'm missing something.

### DIFF
--- a/linkml_runtime/utils/schemaview.py
+++ b/linkml_runtime/utils/schemaview.py
@@ -9,6 +9,7 @@ from collections import defaultdict, deque
 from pathlib import Path, PurePath
 from typing import Mapping, Optional, Tuple, TypeVar
 import warnings
+from urllib.parse import urlparse
 
 from linkml_runtime.utils.namespaces import Namespaces
 from deprecated.classic import deprecated
@@ -18,7 +19,6 @@ from linkml_runtime.utils.pattern import PatternResolver
 from linkml_runtime.linkml_model.meta import *
 from linkml_runtime.exceptions import OrderingError
 from enum import Enum
-from urllib.parse import urlparse
 
 logger = logging.getLogger(__name__)
 
@@ -107,18 +107,32 @@ def is_absolute_path(path: str) -> bool:
     drive, tail = os.path.splitdrive(norm_path)
     return bool(drive and tail)
 
-def resolve_import(sn: str, i: str) -> str:
-    if is_absolute_path(i):
-        return i
-    if urlparse(i).scheme:
-        return i
+#  WINDOWS:
+#     # This cannot be simplified. os.path.normpath() must be called before .as_posix()
+#     i = PurePath(os.path.normpath(PurePath(sn).parent / i)).as_posix()
+#     # i = resolve_import(sn, i)
+# else:
+#     # i = resolve_import(sn, i)
+#     i = os.path.normpath(str(Path(sn).parent / i))
+
+def _resolve_import(source_sch: str, imported_sch: str) -> str:
+    if os.path.isabs(imported_sch):
+        # Absolute import paths are not modified
+        return imported_sch
+    if urlparse(imported_sch).scheme:
+        # File with URL schemes are not modified
+        return imported_sch
+    
+    if WINDOWS:
+        path = PurePath(os.path.normpath(PurePath(source_sch).parent / imported_sch)).as_posix()
     else:
-        path = os.path.normpath(str(Path(sn).parent / i))
-        if i.startswith(".") and not path.startswith("."):
-            # Above condition handles cases where both sn and i are relative paths, and path is not already relative
-            return f"./{path}"
-        else:
-            return path
+        path = os.path.normpath(str(Path(source_sch).parent / imported_sch))
+
+    if imported_sch.startswith(".") and not path.startswith("."):
+        # Above condition handles cases where both source schema and imported schema are relative paths: these should remain relative
+        return f"./{path}"
+
+    return path
 
 
 @dataclass
@@ -305,10 +319,21 @@ class SchemaView(object):
                     # no self imports ;)
                     if i == sn:
                         continue
-                    else:
-                        temp = i
-                        i = resolve_import(sn, i)
-                    _ = sn, temp, i
+
+                    # resolve relative imports relative to the importing schema, rather than the
+                    # origin schema. Imports can be a URI or Curie, and imports from the same
+                    # directory don't require a ./, so if the current (sn) import is a relative
+                    # path, and the target import doesn't have : (as in a curie or a URI)
+                    # we prepend the relative path. This WILL make the key in the `schema_map` not
+                    # equal to the literal text specified in the importing schema, but this is
+                    # essential to sensible deduplication: eg. for
+                    # - main.yaml (imports ./types.yaml, ./subdir/subschema.yaml)
+                    # - types.yaml
+                    # - subdir/subschema.yaml (imports ./types.yaml)
+                    # - subdir/types.yaml
+                    # we should treat the two `types.yaml` as separate schemas from the POV of the
+                    # origin schema.
+                    i = _resolve_import(sn, i)
                     todo.append(i)
 
             # add item to closure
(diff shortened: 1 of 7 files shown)


sneakers-the-rat:
can you add a set of minimal schemas that demonstrates the bug to the tests?

see: https://github.com/linkml/linkml-runtime/blob/a887aec537edde3377e56e2933579958617a66e5/tests/test_utils/test_schemaview.py#L521

https://github.com/linkml/linkml-runtime/tree/main/tests/test_utils/input/imports_relative

sneakers-the-rat:
related to: https://github.com/linkml/linkml-runtime/pull/350

will-0:
@sneakers-the-rat test added that fails with original implementation, and fix that passes all tests.

Noted that some other work has been done to allow relative imports without './' prefix in #350. Can try and reconcile with the above if necessary.

linkml_runtime/utils/schemaview.py
```diff
@@ -104,6 +105,19 @@ def is_absolute_path(path: str) -> bool:
     drive, tail = os.path.splitdrive(norm_path)
     return bool(drive and tail)
 
+def resolve_import(sn: str, i: str) -> str:
```
sneakers-the-rat, line 108: if we're splitting this out to a separate function, it would be great to not have the arguments be named `sn` and `i`

sneakers-the-rat, line 108:
```suggestion
def _resolve_import(sn: str, i: str) -> str:
```

unless we expect this to be used elsewhere, probably make it private. it seems pretty specific to this one case in schemaview

linkml_runtime/utils/schemaview.py
```diff
@@ -104,6 +105,19 @@ def is_absolute_path(path: str) -> bool:
...
+    if urlparse(i).scheme:
+        return i
+    else:
+        path = os.path.normpath(str(Path(sn).parent / i))
+        if i.startswith(".") and not path.startswith("."):
+            # Above condition handles cases where both sn and i are relative paths, and path is not already relative
+            return f"./{path}"
+        else:
+            return path
+
```
sneakers-the-rat, line 120: since we are already treating everything that isn't absolute or has a scheme as a path, i think this would be double-fixing the problem. e.g. tests pass with this

```suggestion
        return os.path.normpath(str(Path(sn).parent / i))

```

sneakers-the-rat (commented):
seems fine, does need to be synchronized with the other changes that are being made to this exact place tho :)

linkml_runtime/utils/schemaview.py
will-0: This was here to retain compatibility with hard-coded paths in `test_imports_relative`. If removed, following error raised:

```
test_utils/test_schemaview.py::test_imports_relative failed: def test_imports_relative():
...
>       assert closure == [
            'linkml:types',
            '../neighborhood_parent',
            'neighbor',
            '../parent',
            '../L1_0_1/L2_0_1_0/grandchild',
            '../../L0_1/L1_1_0/L2_1_0_0/apple',
            '../../L0_1/L1_1_0/L2_1_0_0/index',
            '../../L0_1/L1_1_0/L2_1_0_1/banana',
            '../../L0_1/L1_1_0/L2_1_0_1/index',
            '../../L0_1/L1_1_0/index',
            '../../L0_1/cousin',
            '../L1_0_1/dupe',
            './L2_0_0_0/child',
            './L2_0_0_1/child',
            'main'
        ]
E       AssertionError: assert ['linkml:type...0/apple', ...] == ['linkml:type...0/apple', ...]
E         
E         At index 12 diff: 'L2_0_0_0/child' != './L2_0_0_0/child'
E         Use -v to get more diff
```

linkml_runtime/utils/schemaview.py
will-0: Have renamed to `source_sch` and `imported_sch`

linkml_runtime/utils/schemaview.py
```diff
@@ -289,22 +303,10 @@ def imports_closure(self, imports: bool = True, traverse: Optional[bool] = None,
...
-                    # - types.yaml
-                    # - subdir/subschema.yaml (imports ./types.yaml)
-                    # - subdir/types.yaml
-                    # we should treat the two `types.yaml` as separate schemas from the POV of the
-                    # origin schema.
-                    if sn.startswith('.') and ':' not in i:
-                        i = os.path.normpath(str(Path(sn).parent / i))
+                    else:
+                        temp = i
+                        i = resolve_import(sn, i)
```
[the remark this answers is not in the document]
will-0: Thanks, have included this.

linkml_runtime/utils/schemaview.py
```diff
@@ -289,22 +303,10 @@ def imports_closure(self, imports: bool = True, traverse: Optional[bool] = None,
...
+                    _ = sn, temp, i
```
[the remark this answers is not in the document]
will-0: Nothing, apologies! Was purely for inspection during debugging, meant to remove.

linkml_runtime/utils/schemaview.py
```diff
@@ -289,22 +303,10 @@ def imports_closure(self, imports: bool = True, traverse: Optional[bool] = None,
...
-                    # path, and the target import doesn't have : (as in a curie or a URI)
-                    # we prepend the relative path. This WILL make the key in the `schema_map` not
-                    # equal to the literal text specified in the importing schema, but this is
-                    # essential to sensible deduplication: eg. for
-                    # - main.yaml (imports ./types.yaml, ./subdir/subschema.yaml)
-                    # - types.yaml
-                    # - subdir/subschema.yaml (imports ./types.yaml)
-                    # - subdir/types.yaml
-                    # we should treat the two `types.yaml` as separate schemas from the POV of the
-                    # origin schema.
```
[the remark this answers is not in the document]
will-0: Have put this back in.

amc-corey-cox (approved):
This looks good to me.

amc-corey-cox:
I'm going to wait to merge this until we land #345 to make sure we don't create any merge conflicts.

linkml_runtime/utils/schemaview.py
sneakers-the-rat: ah yes. I do think it's useful to have all paths, internally in the `imports_closure` method, have a `./` or `../` in them so we can determine which need directory traversal trivially, but also acknowledge that would probably be done better with proper typing rather than string conventions. Maybe changing the comment above to something like "ensure any path that requires directory traversal has a ./ prepended" or something like that would be good, but since the above is so anticipatory of future needs (e.g. resolving remote imports, resolving content-ID imports, other contexts where we need to differentiate posix-like directories from other formats) that it borders on YAGNI so i could go either way on what happens with this. thx for explaining

ialarmedalien:
@will-0 Please make sure that you look at other pull requests to ensure that you're not fixing a bug that has already been fixed in another PR.

will-0:
@ialarmedalien certainly, will make sure I do so more thoroughly in the future. Do you have the PR you're referring to? Did search up issues / PRs before but didn't see one at the time.

sneakers-the-rat:
i think they're referring to this, i think they end up being equivalent?
https://github.com/linkml/linkml-runtime/pull/350

PR #369, opened 2025-02-10
Fix test upstream: `-y` to pip uninstall

sneakers-the-rat:
We have a pileup of related PRs:
- https://github.com/linkml/linkml-runtime/pull/360 - currently failing because of this, needs to be merged ASAP
- https://github.com/linkml/linkml-runtime/pull/350 - fails on windows, masked by 360
- https://github.com/linkml/linkml-runtime/pull/367 - includes this, but does more
- https://github.com/linkml/linkml-runtime/pull/364 - also touches this action orthogonally (and removes the need for this) but needs changes in upstream

Trying to break the backlog with this tiny little commit that should let us do 360, which then will let us handle the rest

### DIFF
--- a/.github/workflows/test-upstream.yaml
+++ b/.github/workflows/test-upstream.yaml
@@ -68,7 +68,7 @@ jobs:
       # make extra sure we're removing any old version of linkml-runtime that exists
       - name: uninstall potentially cached linkml-runtime
         working-directory: linkml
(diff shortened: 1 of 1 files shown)


amc-corey-cox (approved):
This seems quite straightforward to me. I'm approving.

PR #376, opened 2025-03-21
Fix "workflow_dispatch" mode for test-upstream action

dalito:
Adds a missing quote which led to failing runs when the job was triggered via "workflow_dispatch" option.


sneakers-the-rat (approved):
oopsie

amc-corey-cox (approved):
Yep, that makes sense.

PR #377, opened 2025-03-24
add isodate as dependency

will-0:
isodate used since [a1a6b48](https://github.com/linkml/linkml-runtime/commit/a1a6b4857b4fd44c1d7199ea1d0ef4a044d2b978), but not added as dependency.


will-0:
Makes sense. How does it work if you're using higher python versions then? My project was failing due to the missed dependency on python 3.12

dalito:
Ah. The import should also depend on the Python version. I should have done that in #357 (surprising that the tests passed). Can you add the conditional import to your PR? (here https://github.com/linkml/linkml-runtime/blob/4f31524e3d6b61d6f7d6f84f761b657603137092/linkml_runtime/utils/metamodelcore.py#L10)

dalito:
@will-0 - FYI: I submitted #380 to solve the issue by making the import Python version dependent.

PR #341, opened 2024-10-01
BF: Bug fixes and other improvement in `SchemaBuilder.add_enum()`

candleindark:
This PR does the following to `SchemaBuilder.add_enum()`.

Bug fixes (tests added):
1. Correct adding of permissible values that are `PermissibleValue` objects
2. Correct handling of `enum_def` values of different types

Other improvements (no test needed):
1. Provide missing documentation of params in the docstring
2. Convert `None` value in `permissible_values` to an empty list

Note: This PR is based on https://github.com/linkml/linkml-runtime/pull/338. To get the relevant commits of this PR, wait for https://github.com/linkml/linkml-runtime/pull/338 is merged to the main branch, and merge the main branch to this branch.

PR #378, opened 2025-03-24
Fix Schemaview.get_uri(...) for no default_prefix in schema

dalito:
upstream_repo: dalito/linkml
upstream_branch: issue2578-fix-uri-in-snapshot

Some test fail with upstream linkml but these tests seem wrong. They test for the wrong uris with a colon, for example [here](https://github.com/linkml/linkml-runtime/actions/runs/14045096004/job/39324014283?pr=378#step:16:926):

```
FAILED tests/test_utils/test_owl.py::test_pred_types - assert value matches snapshot /home/runner/work/linkml-runtime/linkml-runtime/linkml/tests/test_utils/__snapshots__/owl2.owl
  ----- Missing Triples -----
  
  <http://example.org/owl2/:annotslot> a owl:ObjectProperty,
          <https://w3id.org/linkml/SlotDefinition> ;
      rdfs:label "annotslot" ;
      rdfs:domain <http://example.org/owl2/:C1> ;
      rdfs:range <http://example.org/owl2/:C2> ;
      skos:inScheme <http://example.org/owl2> .

(...)

----- Added Triples -----
  
  <http://example.org/owl2/annotslot> a owl:ObjectProperty,
          <https://w3id.org/linkml/SlotDefinition> ;
      rdfs:label "annotslot" ;
      rdfs:domain <http://example.org/owl2/C1> ;
      rdfs:range <http://example.org/owl2/C2> ;
      skos:inScheme <http://example.org/owl2> .
```

This validates the presence of issue linkml/linkml#2578 but is not testing for the correct URIs.
(...or are we wrong in what we assume as correct URIs?)

Closes linkml/linkml#2578


linkml_runtime/utils/schemaview.py
```diff
@@ -1105,13 +1105,18 @@ def get_uri(self, element: Union[ElementName, Element], imports=True, expand=Fal
...
-            pfx = schema.default_prefix
             if use_element_type:
                 e_type = e.class_name.split("_",1)[0]  # for example "class_definition"
                 e_type_path = f"{e_type}/" 
             else:
                 e_type_path = ""
-            uri = f'{pfx}:{e_type_path}{e_name}'
+            pfx = schema.default_prefix
+            # To construct the uri we have to find out if the schema has a default_prefix 
+            # or if a pseudo "prefix" was derived from the schema id.
```
Silvanoc, line 1115: `schema.id` becomes the prefix of a URI and then it's not a "pseudo prefix".

If you consider the term "prefix" _occupied_ by the meaning "CURIE prefix" and have therefore added the "pseudo", then I would simply avoid the term "prefix" in this line. What about `# or if default_prefix is using schema id to build up a URI` or so?

Silvanoc (approved):

dalito:
@Silvanoc This PR has already a test for the issue it solves. Do you think testing with a full schema (#379) provides a stronger safety net for future regressions?

Silvanoc:
> @Silvanoc This PR has already a test for the issue it solves. Do you think testing with a full schema (#379) provides a stronger safety net for future regressions?

Sorry, my bad. I oversaw it.

As long as `SchemaDefinition` does not add a `default_prefix` "automagically", your test would suffice. You have tested that the test fails without your fix, right? Then I can close my PR.

dalito:
Yes, I verified that the test reproduces the issue before adding the fix.

dalito:
@sierra-moxon - #341 was not 3.13 compatible but merged. The introduced typing issues in `test_schema_builder.py` are also fixed in this PR.

ptgolden:
The remaining test failure is also from #341 not being compatible with the changes merged in #345. Specifically the changes in 43b3b3d18970c52e797cc0b74d1fbb8b4da47eb1. That change made it so that the `__init__` function of dataclasses is no longer overridden so that they accept arbitrary args and kwargs. Previously, the expected error was a `ValueError` (raised [here](https://github.com/linkml/linkml-runtime/blob/86f7a4c453253445d956a8e60ad4382cbe3902f3/linkml_runtime/utils/yamlutils.py#L47)).

With that logic no longer in place, the expected error is a `TypeError`, which is what dataclasses normally throw for unexpected args/kwargs.

Pretty sure the fix here is to just expect a `TypeError` here https://github.com/linkml/linkml-runtime/blob/86f7a4c453253445d956a8e60ad4382cbe3902f3/tests/test_utils/test_schema_builder.py#L250

dalito:
I created a PR to fix the test-data-snapshot in linkml (https://github.com/linkml/linkml/pull/2648) and configured this PR to test against that PR. This made the tests pass here.

So this should be good for approval now.

PR #379, opened 2025-03-25
test(schemaview): schema without default_prefix

Silvanoc:
Add regression test for issue #2578 [^1].

Issue was that SchemaView method `get_uri` was adding an errorneous ':' when no `default_prefix` was being declared in the schema.

[^1]: https://github.com/linkml/linkml/issues/2578


Silvanoc:
Already covered by tests added by @dalito in PR #378

PR #381, opened 2025-03-28
recreation of #380 in a branch, with main merged in

sierra-moxon:
upstream_branch: python-3.13

@dalito


dalito:
This is probably no longer needed as it was an experiment to add #380 just before releasing v1.9.1,

#380 passes now all tests (after a rebase on current main).

PR #382, opened 2025-03-28
recreate 378 in branch with updated main Schemaview geturi nofork

sierra-moxon:
upstream_branch: python-3.13


dalito:
I updated the original #378 so this is obsolete.

PR #383, opened 2025-03-28
Cherry-pciked fixes from #378 for rc2

dalito:
upstream_branch: python-3.13

This cherry-picks the two last commits from #378 to fix the typing issue (Dict/dict, List/list) and the type of expected exception in `tests/test_utils/test_schema_builder.py`.

As discussed in dev call the main action is not triggered on merging to main. I added a fix for the trigger.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -4,7 +4,12 @@
 
 name: Build and test linkml-runtime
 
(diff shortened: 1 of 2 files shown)


sierra-moxon (approved):
@dalito - many thanks for pushing this forward!!! 🎉

PR #384, opened 2025-03-30
restore dataclass_extensions. See #345

cmungall:
upstream_repo: linkml/linkml
upstream_branch: pr-2358

This was prematurely removed in #345. A lot of existing code depends on this

Note the changes from #345 temporarily made their way into a 1.9.0 release, but this was promptly yanked


dalito:
#345 also removed the test for this code which was in the file `tests/test_issues/test_issue_718a.py`

cmungall:
This looks like it's failing some of the downstream linkml tests.

I looked at the first one, which dates back to:
https://github.com/biolink/biolinkml/issues/326

The test is

```python
def test_alt_description_2(input_path):
    with pytest.raises(ValueError, match="description must be supplied"):
        YAMLGenerator(input_path("issue_326a.yaml")).serialize(validateonly=True)
```

This is a little rigid. An exception is still raised, but with the changes in #345, the reason/exception type/message changes (since it is now caught at `init` time)

What I am surprised about is how this made it through before..

dalito:
This seems to fail because dynamic versioning has to be installed differently with poetry 2.x

After `pipx install poetry` add `pipx inject poetry poetry-dynamic-versioning` (as in #364)

.github/workflows/test-upstream.yaml
```diff
@@ -118,7 +118,10 @@ jobs:
           python-version: ${{ matrix.python-version }}
 
       - name: Install poetry
-        run: pipx install poetry
+        run: pipx install poetry==2.0.0
```
dalito: Shouldn't this be `pipx install "poetry<2.0"` ?

sierra-moxon:
🍏 📗 🟢 💚

sierra-moxon (approved):
🥗

PR #386, opened 2025-03-31
emit warnings for python 3.13

sierra-moxon:
Two approaches in the commit history of this PR:
1) retain the dataclass_extensions_376.py module contents, switching on python version to preserve pre-1.9.0 release behavior with <3.13 python: https://github.com/linkml/linkml-runtime/pull/386/commits/7c2164cfa6d9d4654d91171ab99e97b9622efb52
2) remove contents of the module just preserve import ability: https://github.com/linkml/linkml-runtime/pull/386/commits/968e3ef9f605d908f01254f2d7725e7e9cfebb1c

happy for feedback on the best approach.  to me, the conservative approach is the first approach.


linkml_runtime/utils/dataclass_extensions_376.py
```diff
@@ -1,30 +1,54 @@
...
-#
-# The purpose of this extension is to enhance the library to allow additional keyword arguments to passed in
-# and then on to the __post_init__ function that can deal with them accordingly
+import warnings
 
-# Beware that there is no promise that signature of the create function will remain consistent
-loc_fn = dataclasses._create_fn
+# Block unsupported Python versions
+if sys.version_info >= (3, 13):
+    raise RuntimeError(
```
sneakers-the-rat: if i read the discussion in slack correctly, we want this to *not* fail for existing code right? 

The monkeypatch that was removed here is functionally equivalent to the standard dataclass behavior, it's just a matter of when the error for extra parameters is raised. so we could just gate the whole module in this check, and in the python 3.13 branch create some dummy function that does nothing but allows the import to proceed, and then we both preserve existing functionality and allow python 3.13 to work

linkml_runtime/utils/dataclass_extensions_376.py
```diff
@@ -1,30 +1,54 @@
...
+        "Python 3.13 removed internal support for `_create_fn`, which this patch depends on.\n\n"
+        "To resolve this:\n"
+        "  • Upgrade your LinkML schema and code to use LinkML >= 1.9.0, which no longer uses this patch\n"
+        "  • Or migrate to Pydantic models for dataclass behavior\n\n"
+    )
 
+# Warn that this extension is deprecated
+warnings.warn(
+    "The LinkML dataclass extension patch is deprecated and will be removed in a future release.\n"
+    "Consider upgrading to LinkML >= 1.9.0 or switching Python class generation to Pydantic.",
```
sneakers-the-rat:
```suggestion
    "The LinkML dataclass extension patch is deprecated and will be removed in v1.10.0.\n"
    "To upgrade, either:\n"
    "- Regenerate this module with linkml >=1.9.0 (preferred)\n"
    "- Remove the import of dataclass_extensions_376 manually, or\n"
    "- Consider switching Python class generation to Pydantic, which will eventually replace python dataclasses.",
```

usually good to give a specific version we're going to remove it in, and just adding more detail on the upgrade path.

sneakers-the-rat (commented):
Was reviewing but the code was updated and now i can't find my comments to delete them, since looks like you addressed one of them! so just leaving the review as a comment since this is something you're actively working on.

Isn't the goal to deprecate this module?

The monkeypatch is equivalent to normal dataclass behavior, just a matter of when the error happens, so we can just remove it, and declare a dummy function that does nothing to allow the imports in the existing downstream generated python dataclass modules to not fail, but then show the deprecation warning.

I think that making two different behaviors (that amount to handling two different error types) depending on your python version seems pretty error prone, and increases tech debt and complexity rather than reduces it.

linkml_runtime/utils/dataclass_extensions_376.py
sneakers-the-rat: removing this since it's outdated!

linkml_runtime/utils/dataclass_extensions_376.py
sierra-moxon: No, I am not ready to set a specific deprecation release, yet.

sierra-moxon:
> Was reviewing but the code was updated and now i can't find my comments to delete them, since looks like you addressed one of them! so just leaving the review as a comment since this is something you're actively working on.
> 
> Isn't the goal to deprecate this module?
> 
> The monkeypatch is equivalent to normal dataclass behavior, just a matter of when the error happens, so we can just remove it, and declare a dummy function that does nothing to allow the imports in the existing downstream generated python dataclass modules to not fail, but then show the deprecation warning.
> 
> I think that making two different behaviors (that amount to handling two different error types) depending on your python version seems pretty error prone, and increases tech debt and complexity rather than reduces it.

Thanks for your feedback, I will take it under review.

linkml_runtime/utils/dataclass_extensions_376.py
sneakers-the-rat: i guess i'm a little unclear on the plan here then - i assumed we would just be marking this as deprecated now and going with the standard dataclass impl, which is functionally equivalent, so i'm not really sure why not set a removal version, but up to you

sneakers-the-rat (approved):

cmungall (changes requested):
The functionality absolutely has to stay!

https://github.com/search?q=%22from+linkml_runtime.utils.dataclass_extensions_376+import+dataclasses_init_fn_with_kwargs%22&type=code

sneakers-the-rat:
ah yeah, looking again, the monkeypatch is applied both at import time of `dataclass_extensions_376` and also within the generated modules themselves. 

so it should be this

```
dataclasses_init_fn_with_kwargs = dataclasses._init_fn    
```

ran the tests in `linkml/main` against that and seems to work fine

linkml_runtime/utils/dataclass_extensions_376.py
sierra-moxon: We need to establish a release process that incorporates checks for deprecation removals as part of each release. Additionally, we should create small, independent pull requests (PRs) to handle these removals. It would be beneficial to coordinate the timing of these removals as an extra bonus (for example, we should ask if folks have bandwidth, assign the removal tasks, and remind folks to get the deprecation PRs in when a release is coming - and further it should be predictable when a release is coming - we lost the plot on this particular bit for 1.9.0 and will need to bump the "removed_in" annotations for a few things).  <-- all of this is in progress at the moment.  

I appreciate the concept and the implementation you’ve developed, and the tests that support it.

sierra-moxon:
Testing Biolink-Model:

python 3.11
in pyproject.toml:
```
...
linkml-runtime = { git = "https://github.com/linkml/linkml-runtime.git", branch = "no_merge_create_fn" }
linkml = ">=1.6.3"
...
```
```

% poetry show linkml
 name         : linkml                             
 version      : 1.8.7                              
 description  : Linked Open Data Modeling Language 

% poetry show linkml-runtime
 name         : linkml-runtime                                                         
 version      : 1.9.0.post9.dev0+968e3ef                                               
 description  : Runtime environment for LinkML, the Linked open data modeling language 

```

% poetry lock --no-update
% rm -rf .venv 
% poetry install
% make test

One of the custom dependencies of `make test` is:

```
id-prefixes:
	$(RUN) gen-python class_prefixes.yaml > src/biolink_model/scripts/classprefixes.py
	cd src/biolink_model/scripts/ && $(RUN) python id_prefixes.py
```

gen-python creates this:
```
# Auto generated from class_prefixes.yaml by pythongen.py version: 0.0.1
# Generation date: 2025-04-01T15:00:59
# Schema: BiolinkClassPrefixes
#
# id: biolink-model-class-prefixes
# description: preferred order identifier prefixes per class in Biolink Model
# license: https://creativecommons.org/publicdomain/zero/1.0/

import dataclasses
import re
from dataclasses import dataclass
from datetime import (
    date,
    datetime,
    time
)
...
```

I get this error:
```
SMoxon@SMoxon-M82 biolink-model % cd src/biolink_model/scripts/ && poetry run python id_prefixes.py
Configuration file exists at /Users/SMoxon/Library/Preferences/pypoetry, reusing this directory.

Consider moving TOML configuration files to /Users/SMoxon/Library/Application Support/pypoetry, as support for the legacy directory will be removed in an upcoming release.
Traceback (most recent call last):
  File "/Users/SMoxon/Documents/src/biolink-model/src/biolink_model/scripts/id_prefixes.py", line 4, in <module>
    from classprefixes import BiolinkClassPrefixMap, Prefix, BiolinkClassPrefixesCollection
  File "/Users/SMoxon/Documents/src/biolink-model/src/biolink_model/scripts/classprefixes.py", line 88, in <module>
    @dataclass(repr=False)
     ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/SMoxon/.pyenv/versions/3.11.0/lib/python3.11/dataclasses.py", line 1211, in wrap
    return _process_class(cls, init, repr, eq, order, unsafe_hash,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/SMoxon/.pyenv/versions/3.11.0/lib/python3.11/dataclasses.py", line 1028, in _process_class
    _init_fn(all_init_fields,
TypeError: 'NoneType' object is not callable
```

I can run this target successfully with the module in place, not just as a placeholder with deprecation warnings, etc... 
```
linkml-runtime = { git = "https://github.com/linkml/linkml-runtime.git", branch = "non-gutted-fix" }
```

With the module removed, and using:
`dataclasses_init_fn_with_kwargs = dataclasses._init_fn` 

```
linkml-runtime = { git = "https://github.com/linkml/linkml-runtime.git", branch = "no_merge_create_fn" }
```
success.

No longer _overriding_ an internal function, but still a bit of debt in referencing the internal function.

sierra-moxon:
I also attempted to run the same test in Biolink repo using python 3.13 venv, but it was blocked by dependency issues not related to this PR. 

```
Installing pydantic-core (2.14.5): Failed
```

trying oak next.

sierra-moxon:
my testing of oak is similar to the testing of biolink for this particular change.

sierra-moxon:
tested nmdc-schema build & test with this branch and python 3.11 and 3.13 -- success. @turbomam

sierra-moxon:
I think I am ready to merge this PR in @cmungall - can you do a quick re-review?

cmungall (approved):

sneakers-the-rat:
> I feel conflicted about making this a DeprecationWarning vs. a FutureWarning but have opted for the less annoying DeprecationWarning in this PR (I am not confident that schema repo managers turn on deprecation warnings periodically in their repos to catch this).

i get the concern, but for the record @sierra-moxon you did the right thing - all test runners should show deprecation warnings, and there is only so much we can do to preserve backwards compat while still maintaining and improving the package, think a fair line to draw is 'if you aren't testing your code regularly or minding the warnings that get emitted during tests, you might experience some deprecation problems.'

PR #387, opened 2025-04-12
Update pyproject.toml & actions to poetry 2.x / PEP 621

dalito:
This is a clean start based on #364


.github/dependabot.yml
```diff
@@ -0,0 +1,9 @@
+# Please see the documentation for all configuration options:
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+- package-ecosystem: github-actions
```
ialarmedalien, line 6: Is there a plan to have dependabot manage python deps, too?

dalito: Not that I am aware of. But I remember that Patrick mentioned another tools for updating the dependencies somewhere else (renovate?). I would suggest to do it step by step (so not in this PR).

ialarmedalien: Dependabot currently supports poetry-mediated python dependency management [via the `pip` configuration](https://docs.github.com/en/code-security/dependabot/ecosystems-supported-by-dependabot/supported-ecosystems-and-repositories#pip-and-pip-compile).

[uv support](https://github.blog/changelog/2025-03-13-dependabot-version-updates-now-support-uv-in-general-availability/) has recently been announced.

I suggest adding in a config section for watching python deps since you're adding in GHA support in this PR.

.github/workflows/pypi-publish.yaml
```diff
@@ -1,35 +1,71 @@
...
-  build-n-publish:
-    name: Build and publish Python 🐍 distributions 📦 to PyPI
+  build:
+    name: Build Python 🐍 distributions 📦 for publishing to PyPI
     runs-on: ubuntu-latest
-
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v4.2
+      # By default the ref (branch, tag or SHA) that triggered the workflow,
```
ialarmedalien, line 15: extremely pedantic :grin: : no need for the comma after "workflow"

.github/workflows/pypi-publish.yaml
```diff
@@ -1,35 +1,71 @@
...
+      run: |
+        pipx install poetry
+        pipx inject poetry poetry-dynamic-versioning
 
     - name: Build source and wheel archives
       run: |
-        poetry version $(git describe --tags --abbrev=0)
         poetry build
+        # Check if correct tag is available
+        git describe --tags --abbrev=0
```
ialarmedalien, line 33: what happens if the correct tag isn't available?

.github/workflows/pypi-publish.yaml
```diff
@@ -1,35 +1,71 @@
...
+          user: __token__
+          password: ${{ secrets.pypi_password }}
+          # verbose: true
+        
+# Used actions: (updates managed by dependabot)
+# - https://github.com/actions/checkout
+# - https://github.com/actions/setup-python
+# - https://github.com/actions/upload-artifact
+# - https://github.com/actions/download-artifact
+# - https://github.com/pypa/gh-action-pypi-publish/
```
ialarmedalien, line 71: 1. Some files have this list at the top, some at the bottom. Can we standardise on a location?

2. Is this list useful if deps are managed by dependabot anyway?

.github/workflows/test-upstream.yaml
```diff
@@ -83,15 +84,15 @@ jobs:
...
           path: linkml
           ref: "${{ env.upstream_branch }}"
           fetch-depth: 0
 
       - name: checkout linkml-runtime
-        uses: actions/checkout@v4
+        uses: actions/checkout@v4.2
         with:
           # don't specify repository like this or else we won't get pull request branches correctly
           # repository: linkml/linkml-runtime
```
ialarmedalien, line 98: this looks like cruft - can it be removed?

.github/workflows/test-upstream.yaml
```diff
@@ -113,22 +114,18 @@ jobs:
...
-      - name: Install dynamic versioning plugin
-        run: poetry self add "poetry-dynamic-versioning[plugin]"
+        run: |
+          pipx install poetry
+          pipx inject poetry poetry-dynamic-versioning
 
       - name: Load cached venv
         id: cached-poetry-dependencies
-        uses: actions/cache@v3
+        uses: actions/cache@1bd1e32a3bdc45362d1e726936510720a7c30a57
```
ialarmedalien, line 128: Why do some actions use specific commits but others use versions? Can we standardise?

pyproject.toml
```diff
@@ -1,67 +1,79 @@
...
     "Programming Language :: Python :: 3.10",
     "Programming Language :: Python :: 3.11",
     "Programming Language :: Python :: 3.12",
     "Programming Language :: Python :: 3.13",
 ]
-
-packages = [
-    { include = "linkml_runtime" }
+include = ["linkml_runtime"]
+requires-python = ">=3.9"
```
ialarmedalien, line 36: do we need to add an upper bound?

.github/workflows/pypi-publish.yaml
dalito: Hmm. I just found out that it is surprisingly possible to create a release without selecting a tag on the release-creation page. So it may be safer to not rely on it and use the full checkout with tags to derive the version safely. I'll change this.

.github/workflows/pypi-publish.yaml
dalito: It is only useful for other developers. Dependabot does not need it. Maybe we should remove this info everywhere?

pyproject.toml
dalito: Not for requires-python, see https://github.com/python-poetry/poetry/issues/10136

.github/workflows/pypi-publish.yaml
```diff
@@ -1,35 +1,71 @@
...
+        uses: actions/download-artifact@v4.2
+        with:
+          name: distribution-files
+          path: dist
+
+      - name: Publish package 📦 to PyPI
+        if: github.event_name == 'release'
+        uses: pypa/gh-action-pypi-publish@v1.12
+        with:
+          user: __token__
```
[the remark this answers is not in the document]
dalito: I just copied the code from the former version.  `__token__` is supplied by the pypi-publish action https://github.com/pypa/gh-action-pypi-publish/?tab=readme-ov-file#specifying-a-different-username 

But based on the docs the line `user: __token__` is not be needed. I remove it.

.github/workflows/check-dependencies.yaml
```diff
@@ -26,17 +30,22 @@ jobs:
       #          install poetry
       #----------------------------------------------
       - name: Install Poetry
-        # Pin to 1.3.2 to workaround https://github.com/python-poetry/poetry/issues/7611
-        run: pipx install poetry==1.3.2
+        run: |
+          pipx install poetry
+          pipx inject poetry poetry-dynamic-versioning
```
ialarmedalien, line 35: don't need the dynamic versioning plugin unless you're releasing a package to pypi

dalito: Your comment made me look at the action in detail: I believe it does not do what it should do. My understanding is that it should run the tests with updated dependencies but it just installs from the current (possibly outdated) lock file. Then it removes the lockfile but does not create an updated one before running the tests. So the tests are exactly the same as when run from main.yaml.

Would you agree that the idea here is to update the lock file first and then test? In this case dynamic-versioning would be needed for the dependency update.

I'll update the logic.

dalito: @sierra-moxon created the action, but accidentally removed the re-installation with updated dependencies in https://github.com/linkml/linkml-runtime/commit/05792437e27fbdd4fec5aaa64c71100af88a1bc5

ialarmedalien: I also noticed that this action seems to duplicate a lot of the `main.yaml` actions so it may be time for a general GHA overhaul. That's probably an action for a different PR, though.

.github/workflows/test-upstream.yaml
dalito: Done.

.github/workflows/test-upstream.yaml
dalito: Done.

ialarmedalien:
LGTM! Unfortunately I don't have write access to be able to approve this for merging, though.

PR #390, opened 2025-05-09
update minimum pydantic version to be 2.0.0

kevinschaper:
Now that linkml is getting ready to fully deprecate pydantic v1, bump the requirement here

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -48,7 +48,7 @@ dependencies = [
     "requests",
     "prefixmaps >=0.1.4",
     "curies >=0.5.4",
-    "pydantic >=1.10.2, <3.0.0",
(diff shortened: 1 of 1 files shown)


.github/workflows/test-upstream.yaml
```diff
@@ -140,7 +140,7 @@ jobs:
       # the cache will still speedup the rest of the installation
       - name: install linkml
         working-directory: linkml
-        run: poetry sync --no-interaction --all-extras
+        run: poetry install --no-interaction --all-extras
```
dalito, line 143: Why did you switch to `poetry install`? 
`poetry sync` avoids the presence of [untracked outdated packages](https://python-poetry.org/docs/cli/#install) (here from cache).

PR #396, opened 2025-05-20
refactor: move modules from utils to types

Silvanoc:
Move modules defining types rather than utilities from `linkml_runtime.utils` to `linkml_runtime.types`.

Part of refactoring proposed in issue https://github.com/linkml/linkml/issues/2371

**Important note**: getting the changes that this PR tries to address effectively working requires a synchronized effort involving the repos `linkml`, `linkml-runtime` and `linkml-model`. Because the Python generator needs to be adapted so that generated classes take the new path and the model needs to be also generated with the new paths. It's questionable if the effort is worth the gain...


Silvanoc:
This PR will make more sense after merging this repo into `linkml` to have a monorepo.

PR #408, opened 2025-07-17
feat(processing): add module to process inlining

Silvanoc:
Trying to address issue 2813 [1] inconsistencies have been identified WRT how object inlining is behaving depending on the values of `inlined`, `inlined_as_list` and the presence/absence of an identifier in the range class. These inconsistencies appear from the fact that no normalization is happening on schema loading (not only in SchemaLoader, but also in SchemaView) and some consumers apply their own logic.

This patch provides a module that should be used on any schema loading (as of now SchemaLoader and SchemaView) to have a common behavior. The code is structured in a way that it covers all potential combinations in an easy to understand manner.

Unit testing is also provided for the module.

[1]: https://github.com/linkml/linkml/issues/2813


Silvanoc:
This PR will make more sense after merging into `linkml` to have a monorepo.

PR #411, opened 2025-07-18
Fix the remote import issue in SchemaViewer and a couple of misc minor bugs

ialarmedalien:
- Fix the remote import issue in SchemaViewer.
- Fix a couple of bugs where the wrong type is reported in an error message.
- Cut down some needlessly verbose sorting code.


linkml_runtime/utils/schemaview.py
```diff
@@ -445,39 +448,20 @@ def ordered(self, elements: ElementDict, ordered_by: OrderedBy | None = None) ->
...
         """
-        ordered_list_of_names = []
-        ordered_elements = {}
-        for c in elements:
-            ordered_list_of_names.append(c)
-        ordered_list_of_names.sort()
-        for name in ordered_list_of_names:
-            ordered_elements[self.get_element(name).name] = self.get_element(name)
-        return ordered_elements
+        return {name: elements[name] for name in sorted(elements.keys())}
```
ialarmedalien, line 456: dictionaries can be relied upon to return their contents in order of insertion, so no need for all the extra gubbins in the original implementation

linkml_runtime/utils/schemaview.py
```diff
@@ -141,6 +141,9 @@ def is_absolute_path(path: str) -> bool:
     """
     if path.startswith("/"):
         return True
+    if "://" in path:
+        return True
```
ialarmedalien, line 145: cause of the remote import issue!

The import code would benefit from being reworked so that the types are worked out early instead of having several different places where the same or similar checks are made.

tests/test_utils/test_schemaview.py
```diff
@@ -307,6 +328,35 @@ def sv_induced_slots() -> SchemaView:
     return SchemaView(schema_str)
 
 
+@pytest.fixture
+def sv_ordering_tests() -> SchemaView:
```
ialarmedalien, line 332: bonus tests added

tests/test_utils/test_schemaview.py
```diff
@@ -697,6 +747,145 @@ def test_uris_without_default_prefix() -> None:
...
+for value in CREATURE_EXPECTED.values():
+    for s_name, el in value.items():
+        if s_name not in CREATURE_EXPECTED["element"]:
+            CREATURE_EXPECTED["element"][s_name] = set()
+        CREATURE_EXPECTED["element"][s_name].update(el)
+
+
+@pytest.mark.parametrize("schema", ["creature_view", "creature_view_remote", "creature_view_local"])
+@pytest.mark.parametrize("entity", CREATURE_EXPECTED.keys())
+def test_creature_schema_entities_with_without_imports(
```
ialarmedalien, line 778: This can replace a number of the existing tests -- I will clean things up in a subsequent PR.

tests/test_utils/test_schemaview.py
dalito: Maybe also test the case where rank is set to float(inf)

sierra-moxon (approved):

PR #413, opened 2025-07-18
Python code fixes by ruff

vladistan:
These were created by running  `ruff`

Most of them are re-arranging imports, fixing whitespace etc.


vladistan:
Looks like the ruff formatting broke one test.   The problem is actually in the implementation,  looks like `splitlines` generates text with dangling whitespace.  Will fix when we address lint compliance of generated files.

ialarmedalien (approved):

PR #418, opened 2025-07-21
Adding __future__type annotations to yaml_loader, loader_root.

ialarmedalien:
I was looking at the files that call the hbreader package and converted them to use the futuristic type annotations whilst I was there.

Also generated the URI to Path / str(Path) maps in the runtime `__init__` file.

### DIFF
--- a/linkml_runtime/loaders/loader_root.py
+++ b/linkml_runtime/loaders/loader_root.py
@@ -1,7 +1,8 @@
+from __future__ import annotations
+
 from abc import ABC, abstractmethod
 from logging import getLogger
-from pathlib import Path
-from typing import Any, Callable, Optional, TextIO, Union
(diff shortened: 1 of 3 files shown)


linkml_runtime/loaders/loader_root.py
```diff
@@ -132,53 +137,52 @@ def loads(
...
                 return target_class(**data_as_dict)
 
             if isinstance(data_as_dict, JsonObj):
                 return [target_class(**as_dict(x)) for x in data_as_dict]
-            else:
-                raise ValueError(f"Unexpected type {data_as_dict}")
-        else:
-            return None
+
+            msg = f"Unexpected type {data_as_dict}"
```
ialarmedalien, line 158: separate error message construction

removed surplus `else` statements to simplify code

PR #423, opened 2025-07-30
added new variable: element_type in yamlutils.as_json_object() and JSONDumper.dumps()

puja-trivedi:
In attempt to solve [issue#2849](https://github.com/linkml/linkml/issues/2849)

## Proposed Solution:
added new variable: `element_type` that stores the type of `element` prior to it being serialized to a dictionary by `model_dump()` in `JSONDumper.dumps(`). This will allow `@type` in `yamlutils.as_json_object()` to be assigned to the **true** type of `element`, rather than the being assigned to `dict` which is the type returned from `element.model_dump()`. This change only effects cases when `element` is of type `pydantic.BaseModel`.

## Describe the bug
In `linkml_runtime/dumpers/json_dumper.py`, the `dumps()` method calls` element.model_dump()` early in the process, which converts the Pydantic model to a dictionary. This causes the loss of the original object type information needed for proper @type injection.

**Expected Behavior**

When `inject_type=True`, the `@type` field should contain the actual class name of the original object (e.g., `"OrganismTaxon"`, `"Gene"`, etc.).

**Current Behavior**

The `@type` field is set to `"dict"` because the type information is extracted after `model_dump()` converts the object to a dictionary.

### DIFF
--- a/linkml_runtime/dumpers/json_dumper.py
+++ b/linkml_runtime/dumpers/json_dumper.py
@@ -16,7 +16,11 @@ from linkml_runtime.utils.yamlutils import YAMLRoot, as_json_object
 
 class JSONDumper(Dumper):
     def dump(
-        self, element: Union[BaseModel, YAMLRoot], to_file: str, contexts: CONTEXTS_PARAM_TYPE = None, **kwargs
+        self,
+        element: Union[BaseModel, YAMLRoot],
+        to_file: str,
+        contexts: CONTEXTS_PARAM_TYPE = None,
+        **kwargs,
     ) -> None:
         """
         Write element as json to to_file
@@ -34,7 +38,12 @@ class JSONDumper(Dumper):
             element = element.model_dump()
         super().dump(element, to_file, contexts=contexts, **kwargs)
 
-    def dumps(self, element: Union[BaseModel, YAMLRoot], contexts: CONTEXTS_PARAM_TYPE = None, inject_type=True) -> str:
+    def dumps(
+        self,
+        element: Union[BaseModel, YAMLRoot],
+        contexts: CONTEXTS_PARAM_TYPE = None,
+        inject_type=True,
+    ) -> str:
         """
         Return element as a JSON or a JSON-LD string
         :param element: LinkML object to be emitted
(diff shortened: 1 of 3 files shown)


kevinschaper:
```suggestion
        return json.dumps(
            as_json_object(element, contexts, inject_type=inject_type, element_type=element_type),
            default=default,
            ensure_ascii=False,
            indent="  ",
        )
```

kevinschaper:
```suggestion
def as_json_object(
    element: YAMLRoot, contexts: CONTEXTS_PARAM_TYPE = None, inject_type=True, element_type=None
) -> JsonObj:
```

kevinschaper (commented):
I tried to get ai buddy to push the formatting updates as code suggestions, and ended up needing to do it manually, sorry for the PR noise!

linkml_runtime/dumpers/json_dumper.py
```diff
@@ -62,10 +62,11 @@ def default(o):
             else:
                 return json.JSONDecoder().decode(o)
 
+        element_type = element.__class__.__name__
         if isinstance(element, BaseModel):
             element = element.model_dump()
         return json.dumps(
-            as_json_object(element, contexts, inject_type=inject_type), default=default, ensure_ascii=False, indent="  "
+            as_json_object(element, contexts, inject_type=inject_type, element_type=element_type), default=default, ensure_ascii=False, indent="  "
```
kevinschaper, line 69:
```suggestion
        return json.dumps(
            as_json_object(element, contexts, inject_type=inject_type, element_type=element_type),
            default=default,
            ensure_ascii=False,
            indent="  ",
        )
```

puja-trivedi:
@kevinschaper - thank you for your suggestions, all the tests pass now! Let me know if were able to find someone else with jsonld experience that would be able to review this PR.

kevinschaper:
@puja-trivedi This looks great - at first I didn't see why this was working, but it makes sense to me now that it captures the class name while its still a class, and then passes it forward to the json dumper. Could you make a quick unit test that instantiates a pydantic instance, dumps it to json, and confirms that the @type is correct?

kevinschaper (approved):
Perfect! Thank you for fixing this!

PR #446, opened 2025-08-02
Convert test_loaders_pydantic.py to pytest format

vladistan:
- Extract loader_test helper function from base class
- Convert 4 test methods to standalone functions
- Test both YAML and JSON loaders with different models

### DIFF
--- a/tests/test_loaders_dumpers/input/book_series_lotr_yaml.yaml
+++ b/tests/test_loaders_dumpers/input/book_series_lotr_yaml.yaml
@@ -1,0 +1,20 @@
+id: S001
+name: Lord of the Rings
+genres:
+- fantasy
+creator:
+  name: JRR Tolkien
+  from_country: England
+books:
+- id: S001.1
+  name: Fellowship of the Ring
+  summary: Hobbits
+  price: 5.99
+- id: S001.2
+  name: The Two Towers
+  summary: More hobbits
+  price: 5.99
+- id: S001.3
+  name: Return of the King
+  summary: Yet more hobbits
+  price: 6.99
--- a/tests/test_loaders_dumpers/input/kitchen_sink_normalized_inst_01_json.yaml
+++ b/tests/test_loaders_dumpers/input/kitchen_sink_normalized_inst_01_json.yaml
@@ -1,0 +1,41 @@
+persons:
+- id: P:001
+  name: fred bloggs
+  age_in_years: 33
+- id: P:002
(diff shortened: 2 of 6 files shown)


tests/test_loaders_dumpers/test_loaders_pydantic.py
```diff
@@ -1,29 +1,57 @@
...
+    "filename,model,loader",
+    [
+        ("book_series_lotr.yaml", BookSeries, yaml_loader),
+        ("book_series_lotr.json", BookSeries, json_loader),
+        ("kitchen_sink_normalized_inst_01.yaml", Dataset, yaml_loader),
+        ("kitchen_sink_normalized_inst_01.json", Dataset, json_loader),
+    ],
+)
+def test_loader(filename, model, loader):
+    loader_test(filename, model, loader)
```
ialarmedalien, line 57: Do you need to have `loader_test` as a separate function or can you just parametrize it directly (with the appropriate test name, of course 🙂 )?

vladistan: You right...  fixed..

ialarmedalien (dismissed):

vladistan:
weird... works on my machine :-(

vladistan:
Ahh ... thanks for digging it up.    So looks like we are indeed reading other tests outputs,  I don't think it's a right thing to do.  I copied those outputs into the input directory they are rather short.   I think original test was working because it was importing things from other test files which is also not right thing to do.   I am going to clean it up and get ready for merging.

vladistan:
Ok now all tests are passing

sierra-moxon (approved):

tests/test_loaders_dumpers/input/book_series_lotr_yaml.yaml
```diff
@@ -0,0 +1,20 @@
+id: S001
```
ialarmedalien, line 1: This is identical to the other file -- I would just have a single file, `book_series_lotr.yaml`.

tests/test_loaders_dumpers/input/kitchen_sink_normalized_inst_01_json.yaml
```diff
@@ -0,0 +1,41 @@
+persons:
```
ialarmedalien, line 1: it looks like this is also identical to the yaml version -- just add one, no need for duplicates.

PR #450, opened 2025-09-03
Use full versions specs for all github actions

vladistan:


dalito:
Thanks for doing this. I was about to do the same. There are still some conflict markers in the files so I cannot yet approve.

PR #454, opened 2025-09-25
test_schemaview.py: add comprehensive tests for range determination

ialarmedalien:
Added a comprehensive set of tests for testing `slot.range` and `slot.slot_range_as_union`.

There is a basic schema that tests different "types" of range (a type as a range, a class as a range, an enum as a range, a linkml:Any range). In addition to testing this range schema with or without a `default_range` being set in the schema, the range schema is also imported into another schema (with/without default range set), and that schema is then imported into yet another schema (with/without default range set). These various schemas are generated dynamically by the function `sv_range_import_whatever` and a tuple representing how many levels of import there are and what each schema has as its default_range.

### DIFF
--- a/tests/test_utils/test_schemaview.py
+++ b/tests/test_utils/test_schemaview.py
@@ -1019,7 +1019,8 @@ def test_schemaview_enums(schema_view_no_imports: SchemaView) -> None:
 
     There are three schemas used in these tests:
 
-    The `range_local` (RL) schema        - based on RANGE_DATA["schema_path"]["range_local"]
+    The `range_local` (RL) schema
+        - based on RANGE_DATA["schema_path"]["range_local"]
         - the default_range value in this schema is referred to as local_default;
 
     The `range_importer` (RI) schema, which imports the range_local schema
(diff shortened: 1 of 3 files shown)


tests/test_utils/test_schemaview.py
```diff
@@ -1012,6 +1013,403 @@ def test_schemaview_enums(schema_view_no_imports: SchemaView) -> None:
...
+    :return: path to the generated file
+    :rtype: Path
+    """
+    tmp_file = tmp_path_factory.mktemp("test_schemaview") / filename
+    tmp_file.write_text(contents)
+    return tmp_file
+
+
+def gen_schema_name(range_tuple: tuple[str, str | None, str | None]) -> str | None:
+    """Generate a schema name from a range tuple.
```
ialarmedalien, line 1097: This naming schema allows the schema configuration and defaults to be deduced from the name alone. Wow!

tests/test_utils/test_schemaview.py
```diff
@@ -1012,6 +1013,403 @@ def test_schemaview_enums(schema_view_no_imports: SchemaView) -> None:
...
+    :param range_tuple: the default_range for the range schema; if the value of local_default is "", the default_range is not set.
+    :type range_tuple: tuple[str, str|None, str|None]
+    :return: a loaded SchemaView for the appropriate schema(s)
+    :rtype: SchemaView
+    """
+    (local_default, importer_default, import_importer_default) = range_tuple
+
+    if importer_default is None and import_importer_default is not None:
+        err_msg = "If the importer is not used, any import_importer value other than None is not valid"
+        raise ValueError(err_msg)
```
ialarmedalien, line 1192: This will never happen for any of the RANGE_TUPLES combinations, but always worth adding error handling _just in case_.

tests/test_utils/test_schemaview.py
```diff
@@ -1012,6 +1013,403 @@ def test_schemaview_enums(schema_view_no_imports: SchemaView) -> None:
...
+        if importer_default is None:
+            sv = SchemaView(range_local_path)
+        else:
+            # importing the range_local schema using the range_importer schema
+            sv = SchemaView(range_importer_path, importmap=importmap)
+        return check_generated_schemaview(sv, range_tuple)
+
+    # schema that imports the range_importer schema (just specified as text, no need to save it)
+    schema_text = f"id: https://example.com/{RII}\nname: {RII}\n"
+    schema_text += f"\nimports:\n  - {RI}\n\n"
```
ialarmedalien, line 1221: Could probably have put these on the same line. Oh well.

PR #457, opened 2025-09-30
module_name parameter for compile_python resolves linkml/linkml#2903

tfliss:
This change adds a module_name keyword to the compile_python utility method to override the default 'test'.
This makes it possible to compile multiple modules and import one from the other.

An issue test is also included that verifies new behavior and the existing default behavior.

### DIFF
--- a/linkml_runtime/utils/compile_python.py
+++ b/linkml_runtime/utils/compile_python.py
@@ -16,19 +16,21 @@ def file_text(txt_or_fname: str) -> str:
     return txt_or_fname
 
 
-def compile_python(text_or_fn: str, package_path: str = None) -> ModuleType:
+def compile_python(text_or_fn: str, package_path: str | None = None, module_name: str | None = None) -> ModuleType:
     """
     Compile the text or file and return the resulting module
     @param text_or_fn: Python text or file name that references python file
-    @param package_path: Root package path.  If omitted and we've got a python file, the package is the containing
-    directory
+    @param package_path: Root package path.  If omitted and we've got a python file, the package is the containing directory
+    @param module_name: to be used in an import statement, default 'test'
     @return: Compiled module
     """
+    if module_name is None:
+        module_name = "test"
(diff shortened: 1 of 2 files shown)


tests/test_utils/test_compile_python.py
```diff
@@ -0,0 +1,39 @@
+import pytest
+from linkml_runtime.utils.compile_python import compile_python
```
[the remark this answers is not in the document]
tfliss: Great suggestion, I moved it. Unfortunately I don't really have the bandwidth at the moment to add additional tests so the coverage probably isn't adequate for the existing branches.

linkml_runtime/utils/compile_python.py
```diff
@@ -7,28 +7,32 @@
...
+    :return: File content as string
     """
     if len(txt_or_fname) > 4 and "\n" not in txt_or_fname:
         with open(txt_or_fname) as ef:
             return ef.read()
     return txt_or_fname
 
 
-def compile_python(text_or_fn: str, package_path: str = None) -> ModuleType:
+def compile_python(text_or_fn: str, package_path: str = None, module_name: str = None) -> ModuleType:
```
ialarmedalien: both these `str = None` type declarations should be `str | None = None`

linkml_runtime/utils/compile_python.py
```diff
@@ -7,28 +7,32 @@
...
-    directory
-    @return: Compiled module
+
+    :param text_or_fn: Python text or file name that references python file
+    :param package_path: Root package path. If omitted and we've got a python file, the package is the containing directory
+    :param module_name: Used in an import statement, default 'test'
+    :return: Compiled module
     """
+    if module_name is None:
+        module_name = "test"
```
ialarmedalien: suggest using `if not module_name` instead of None -- otherwise, it's possible to create a module with an empty string as the name

PR #464, opened 2025-10-02
schemaview.py: add tests for get_identifier_slot and get_key_slot

ialarmedalien:
Add tests for `get_identifier_slot` and `get_key_slot`.

The method of generating the tests may look suspiciously familiar...

Ideally when linkml and linkml-runtime are merged into a single happy family, some of these repeated fixtures can be eliminated.

### DIFF
--- a/linkml_runtime/utils/schemaview.py
+++ b/linkml_runtime/utils/schemaview.py
@@ -1452,31 +1452,6 @@ class SchemaView:
         return {k: v.value for k, v in e.annotations.items()}
 
     @lru_cache(None)
-    def class_slots(
-        self, class_name: CLASS_NAME, imports: bool = True, direct: bool = False, attributes: bool = True
-    ) -> list[SlotDefinitionName]:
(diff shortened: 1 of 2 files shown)


linkml_runtime/utils/schemaview.py
```diff
@@ -1651,7 +1651,7 @@ def get_identifier_slot(self, cn: CLASS_NAME, use_key: bool = False, imports: bo
         :param imports:
         :return: slot that acts as identifier, if present
         """
-        for sn in self.class_slots(cn, imports=imports):
+        for sn in sorted(self.class_slots(cn, imports=imports)):
```
ialarmedalien, line 1654: Allows the `identifier` (or `key`) slot to be predicted in cases where there are several slots marked as ID/key. Of course, the user will already have run the linter to eliminate the possibility of there being several ID/key slots...

dalito (approved):
This was easy 😄

PR #467, opened 2025-10-07
schemaview.py: make `permissible_value_*` act consistently with other such methods

ialarmedalien:
Two changes:

- Throw an error if the permissible value supplied in the query is _not_ a PV of the enum.

- Change the functioning of the `permissible_value_parent` method to return an empty list if the PV has no parents, which is what other `*_parents` (`class_parents`, `type_parents`, `slot_parents`, etc.) do. It was previously returning `None`.

Note: the naming of this method is inconsistent with the other parent terms and should be fixed at some point.

### DIFF
--- a/linkml_runtime/utils/schemaview.py
+++ b/linkml_runtime/utils/schemaview.py
@@ -820,41 +820,37 @@ class SchemaView:
     @lru_cache(None)
     def permissible_value_parent(
         self, permissible_value: str, enum_name: ENUM_NAME
-    ) -> list[str | PermissibleValueText] | None:
+    ) -> list[str | PermissibleValueText]:
         """:param enum_name: child enum name
         :param permissible_value: permissible value
         :return: all direct parent enum names (is_a)
         """
         enum = self.get_enum(enum_name, strict=True)
-        if enum:
-            if permissible_value in enum.permissible_values:
-                pv = enum.permissible_values[permissible_value]
-                if pv.is_a:
-                    return [pv.is_a]
-            return None
-        return []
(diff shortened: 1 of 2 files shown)


linkml_runtime/utils/schemaview.py
```diff
@@ -820,41 +820,37 @@ def enum_parents(
...
     def permissible_value_parent(
         self, permissible_value: str, enum_name: ENUM_NAME
-    ) -> list[str | PermissibleValueText] | None:
+    ) -> list[str | PermissibleValueText]:
         """:param enum_name: child enum name
         :param permissible_value: permissible value
         :return: all direct parent enum names (is_a)
         """
         enum = self.get_enum(enum_name, strict=True)
-        if enum:
```
ialarmedalien, line 829: `get_enum(enum_name, strict=True)` throws an error if `enum` is not found, so checking for `enum` is unnecessary

linkml_runtime/utils/schemaview.py
```diff
@@ -820,41 +820,37 @@ def enum_parents(
...
-        raise ValueError(msg)
+        pv = enum.permissible_values.get(permissible_value)
+        if not pv:
+            err_msg = f'"{permissible_value}" is not a permissible value of the enum "{enum_name}".'
+            raise ValueError(err_msg)
+
+        return [
+            isa_pv
+            for isa_pv, isa_pv_entity in enum.permissible_values.items()
+            if isa_pv_entity.is_a and pv.text == isa_pv_entity.is_a
```
ialarmedalien, line 852: list comprehension does the same thing as the `for` loop in the original

